PR #7, opened 2021-03-30
feat: expose Dialer and add DialerOptions

kurtisvg:
Exports a Dialer object that can be used to configure the Dialer behavior.

### DIFF
--- a/dial.go
+++ b/dial.go
@@ -22,26 +22,26 @@ import (
 )
 
 var (
-	once sync.Once
-	dm   *Dialer
-	dErr error
+	once          sync.Once
+	defaultDialer *Dialer
+	dErr          error
 )
 
 // Dial returns a net.Conn connected to the specified Cloud SQL instance. The instance argument must be the
 // instance's connection name, which is in the format "project-name:region:instance-name".
 func Dial(ctx context.Context, instance string) (net.Conn, error) {
-	d, err := defaultDialer()
+	d, err := getDefaultDialer()
 	if err != nil {
 		return nil, err
 	}
 	return d.Dial(ctx, instance)
 }
 
-// defaultDialer provides the singleton dialer as a default for dial functions.
-func defaultDialer() (*Dialer, error) {
+// getDefaultDialer provides the singleton dialer as a default for dial functions.
+func getDefaultDialer() (*Dialer, error) {
 	// TODO: Provide functionality for customizing/setting the default dialer
 	once.Do(func() {
-		dm, dErr = NewDialer(context.Background())
+		defaultDialer, dErr = NewDialer(context.Background())
 	})
-	return dm, dErr
+	return defaultDialer, dErr
 }
--- a/dial.go
+++ b/dial.go
@@ -17,19 +17,13 @@ package cloudsqlconn
 
 import (
 	"context"
-	"crypto/rand"
-	"crypto/rsa"
-	"fmt"
 	"net"
 	"sync"
-
-	"cloud.google.com/cloudsqlconn/internal/cloudsql"
-	sqladmin "google.golang.org/api/sqladmin/v1beta4"
 )
 
 var (
 	once sync.Once
-	dm   *dialManager
+	dm   *Dialer
 	dErr error
 )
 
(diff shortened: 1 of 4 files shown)


dial.go
```diff
@@ -40,72 +34,14 @@ func Dial(ctx context.Context, instance string) (net.Conn, error) {
...
-func defaultDialer() (*dialManager, error) {
+func defaultDialer() (*Dialer, error) {
 	// TODO: Provide functionality for customizing/setting the default dialer
 	once.Do(func() {
-		dm, dErr = newDialManager()
+		dm, dErr = NewDialer(context.Background())
 	})
 	return dm, dErr
 }
-
```
enocom, line 54: This is probably stating the obvious, but this code could have easily stayed in this file. Same with options.go below. Typically Go files are larger (since they don't need to map one to one to a class in Java, for example). Mostly it's a matter of convenience to the reader (looking at one file vs jumping between files). Ultimately this is a matter of preference and up to you. Just wanted to call it out.

options.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"golang.org/x/oauth2"
+	apiopt "google.golang.org/api/option"
+)
+
+// A DialerOption is an option for configuring a Dialer.
+type DialerOption func(d *dialerConfig)
+
+// DialerOptions turns a list of DialerOption instances into an DialerOption.
```
enocom, line 25: s/an DialerOption/a DialerOption/

dialer.go
```diff
@@ -0,0 +1,99 @@
...
+
+// NewDialer creates a new Dialer.
+func NewDialer(ctx context.Context, opts ...DialerOption) (*Dialer, error) {
+	// TODO: Add shared / async key generation
+	key, err := rsa.GenerateKey(rand.Reader, 2048)
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate rsa keys: %v", err)
+	}
+
+	var cfg dialerConfig
```
enocom, line 51: Nit: I tend to see the `cfg := &dialerConfig{}` pattern more often when `cfg` is used as a pointer afterwords.

options.go
```diff
@@ -0,0 +1,53 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cloudsqlconn
+
+import (
+	"golang.org/x/oauth2"
+	apiopt "google.golang.org/api/option"
```
enocom, line 19: Why the named import here?

options.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+// WithCredentialsFile returns a DialerOption that specifies a service account or refresh token JSON credentials file to be used as the basis for authentication.
+func WithCredentialsFile(filename string) DialerOption {
+	return func(d *dialerConfig) {
+		d.sqladminOpts = append(d.sqladminOpts, apiopt.WithCredentialsFile(filename))
+	}
+}
+
+// WithCredentialsFile returns a DialerOption that specifies a service account or refresh token JSON credentials to be used as the basis for authentication.
```
enocom, line 41: This docstring is identical to the one above. Maybe we just crib from the options package for these functions since they're basically call-throughs?

dial.go
kurtisvg: If you prefer, I can put them back in one file. I typically find large go files both difficult to navigate and work in, and prefer some organizational boundaries so it's easier to read stuff with proper context. 

These are a bit on the small side at the moment, but I expect them to grow a fair amount so I just split them now.

options.go
kurtisvg: The package name is just "option", which feels non-specific in a folder called `options.go` that (will) include a `DialOption` and `DialerOption` - so it felt more readable to clarify the package name a bit.

dial.go
enocom: Keeping them as is works for me. Just wanted to bring it up.

dial.go
```diff
@@ -17,95 +17,31 @@ package cloudsqlconn
...
 		return nil, err
 	}
-	return d.dial(ctx, instance)
+	return d.Dial(ctx, instance)
 }
 
-// defaultDialer provides the singleton dialer as a default for dial functions.
-func defaultDialer() (*dialManager, error) {
+// getDefaultDialer provides the singleton dialer as a default for dial functions.
+func getDefaultDialer() (*Dialer, error) {
```
enocom, line 41: FYI: https://golang.org/doc/effective_go#Getters

enocom (approved):

PR #11, opened 2021-04-16
fix: perform refresh operations asynchronously

kurtisvg:


internal/cloudsql/instance.go
```diff
@@ -170,15 +170,51 @@ func performRefresh(client *sqladmin.Service, cn connName, k *rsa.PrivateKey, d
 	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
 	defer cancel()
 
-	// TODO: perform these steps asynchronously and return the results
-	md, err := fetchMetadata(ctx, client, cn)
-	if err != nil {
-		return md, nil, fmt.Errorf("fetch metadata failed: %w", err)
+	// start async fetching the instance's metadata
+	type mdRes struct {
```
enocom, line 174: Probably not a big deal, but because this struct is inline, I *think* it will be allocated on each call to `performRefresh`. Any reason not to declare it outside the function?

enocom (commented):
One small question. Nice use of buffered channels.

internal/cloudsql/instance.go
kurtisvg: My understanding is that the Go compiler will treat this so it's only a difference of scope, but I'll double check. 

The only reason was one of scope - it's not used outside of the function so I didn't really want to cloud the namespace with it.

enocom (approved):

internal/cloudsql/instance.go
kurtisvg: SO seems to think there is no performance difference: https://stackoverflow.com/a/42010880/9414803

But I can't seem to find anything more official. I know that it doesn't for anon functions, so it seems likely we're okay here.

PR #18, opened 2021-05-07
fix: rate limit refresh attempts per instance

### LINKED ISSUE #14: Rate limit SQL Admin API requests
Rate limit how often SQL Admin API calls can be triggered

kurtisvg:
Fixes #14


internal/cloudsql/instance.go
```diff
@@ -117,6 +121,7 @@ func NewInstance(instance string, client *sqladmin.Service, key *rsa.PrivateKey,
 	}
 	i := &Instance{
 		connName:       cn,
+		clientLimiter:  rate.NewLimiter(rate.Every(30*time.Second), 2),
```
enocom, line 124: What's the significance of `2` here? Without it, the rate limiter allows a single call every 30 seconds. With it, clients could at most make two calls every 30 seconds. Is that right?

internal/cloudsql/instance.go
```diff
@@ -169,7 +175,13 @@ func (i *Instance) scheduleRefresh(d time.Duration) *refreshResult {
 }
 
 // performRefresh immediately performs a full refresh operation using the Cloud SQL Admin API.
-func performRefresh(ctx context.Context, client *sqladmin.Service, cn connName, k *rsa.PrivateKey) (metadata, *tls.Config, error) {
+func performRefresh(ctx context.Context, client *sqladmin.Service, l *rate.Limiter, cn connName, k *rsa.PrivateKey) (metadata, *tls.Config, error) {
+	// avoid refreshing too often to try not to tax the SQL Admin API quotas
+	err := l.Wait(ctx)
+	if err != nil {
+		return metadata{}, nil, fmt.Errorf("refresh was throttled until context expired: %s", err)
```
enocom, line 182: nit: errors are conventionally rendered with %v.

internal/cloudsql/instance.go
kurtisvg: It's a leaky bucket limiter, with a capacity of 2. So essentially you can "burst" two requests immediately, but will never exceed an average rate of 1 per 30s outside of that. It means that if a one off error occurs, we can instantly try again. However if we consistently have errors, we won't exceed the 1/30s number. 

https://en.wikipedia.org/wiki/Leaky_bucket

enocom (approved):

PR #19, opened 2021-05-07
feat: force early refresh of instance info if connect fails

### LINKED ISSUE #13: Trigger refresh on connection failures
If a connection fails to connect, attempt a refresh operation to make sure the info is up to date.

kurtisvg:
Fixes #13

### DIFF
--- a/internal/cloudsql/instance.go
+++ b/internal/cloudsql/instance.go
@@ -147,6 +147,18 @@ func (i *Instance) ConnectInfo(ctx context.Context) (map[string]string, *tls.Con
 	return res.md.ipAddrs, res.tlsCfg, nil
 }
 
+// ForceRefresh triggers an immediate refresh operation to be scheduled and used for future connection attempts.
+func (i *Instance) ForceRefresh() {
+	i.resultGuard.Lock()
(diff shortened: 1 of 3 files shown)


internal/cloudsql/instance.go
```diff
@@ -147,6 +147,18 @@ func (i *Instance) ConnectInfo(ctx context.Context) (map[string]string, *tls.Con
 	return res.md.ipAddrs, res.tlsCfg, nil
 }
 
+// ForceRefresh triggers an immediate refresh operation to be scheduled and used for future connection attempts.
+func (i *Instance) ForceRefresh() {
+	i.resultGuard.Lock()
+	defer i.resultGuard.Unlock()
+	// If the next refresh hasn't started yet, we can cancel it and start an immediate one
+	if i.next.Cancel() {
```
enocom, line 155: Slightly tangential: I see that we're saving reference to `i.timer` only to stop it here, but we're not draining the timer's channel when stopping it. Since we're not using the channel, this isn't a big deal (I think), but still wonder: should we drain the channel in `Cancel`?

enocom (approved):

Issue #9, opened 2021-04-09
Add better tests and improve Code Coverage


kurtisvg:

enocom:
- Add an explicit workflow for just code coverage, and make sure it's above X%
- Add a separate step we can output what lines are and are not covered in the logs so we don't always need to run it ourselves too

PR #22, opened 2021-05-09
chore(tests): Use a mock API for testing 

kurtisvg:


internal/cloudsql/refresh.go
```diff
@@ -67,7 +67,7 @@ func fetchMetadata(ctx context.Context, client *sqladmin.Service, inst connName)
 		}
 	}
 	if len(ipAddrs) == 0 {
-		return metadata{}, fmt.Errorf("unsupported instance - contains no valid ip addresses")
+		return metadata{}, fmt.Errorf("cannot connect to instance - it has no supported ip addresses")
```
enocom, line 70: s/ip/IP/

internal/cloudsql/instance_test.go
```diff
@@ -74,7 +92,7 @@ func TestConnectInfo(t *testing.T) {
 		t.Fatalf("failed to generate keys: %v", err)
 	}
 
-	im, err := NewInstance(instConnName, client, key, 30*time.Second)
+	im, err := NewInstance(cn.String(), client, key, 30*time.Second)
```
enocom, line 95: I'd like to improve the assertions in all these tests, but consider that work to be part of #9, especially now that we have this mock API.

internal/mock/cloudsql.go
```diff
@@ -0,0 +1,82 @@
...
+	privKey, cert, err := generateInstanceCerts()
+	if err != nil {
+		return CloudSQLInst{}, err
+	}
+
+	c := CloudSQLInst{
+		project:   project,
+		region:    region,
+		name:      name,
+		dbVersion: "POSTGRES_12", // default of no particular importance
```
enocom, line 52: If this dbVersion doesn't matter, shall we just hard-code an empty string in the mock responses?

internal/mock/sqladmin.go
```diff
@@ -0,0 +1,231 @@
...
+	sync.Mutex
+
+	reqMethod string
+	reqPath   string
+	reqCt     int
+
+	handle func(resp http.ResponseWriter, req *http.Request)
+}
+
+// matches returns true if a given http.Request should be handled by this MockRequest.
```
enocom, line 84: s/MockRequest/Request/

internal/mock/sqladmin.go
```diff
@@ -0,0 +1,231 @@
...
+		return false
+	}
+	if r.reqCt <= 0 {
+		return false
+	}
+	r.reqCt--
+	return true
+}
+
+// InstanceGetSuccess returns a MockRequest that responds to the `instance.get` SQLAdmin
```
enocom, line 101: s/MockRequest/Request/

internal/mock/sqladmin.go
```diff
@@ -0,0 +1,231 @@
...
+				return
+			}
+			resp.WriteHeader(http.StatusOK)
+			resp.Write(b)
+		},
+	}
+	return r
+}
+
+// CreateEphemeralSuccess returns a MockRequest that responds to the
```
enocom, line 151: ditto above.

internal/mock/sqladmin.go
```diff
@@ -0,0 +1,231 @@
...
+				http.Error(resp, fmt.Errorf("unable to decode PublicKey: %w", err).Error(), http.StatusBadRequest)
+				return
+			}
+			pubKey, err := x509.ParsePKIXPublicKey(bl.Bytes)
+			if err != nil {
+				http.Error(resp, fmt.Errorf("unable to decode PublicKey: %w", err).Error(), http.StatusBadRequest)
+				return
+			}
+
+			// Create a signed cert from the requests public key.
```
enocom, line 187: request's

internal/mock/sqladmin.go
```diff
@@ -0,0 +1,231 @@
...
+			cert := &x509.Certificate{ // TODO: Validate this format vs API
```
enocom, line 188: How much of these data is actually necessary for the tests? Could we omit any of it?

internal/mock/sqladmin.go
```diff
@@ -0,0 +1,231 @@
...
+
+	return s.Client(), s.URL, cleanup
+
+}
+
+// Request represents a HTTP request for a test Server to mock responses for.
+//
+// Use NewRequest to initialize new Requests.
+type Request struct {
+	sync.Mutex
```
enocom, line 75: What is this mutex protecting? Could we put it directly above the field(s) is protects?

internal/mock/cloudsql.go
```diff
@@ -0,0 +1,82 @@
...
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"math/big"
+	"time"
+)
+
+// CloudSQLInst represents settings for a specific Cloud SQL instance.
+//
+// Use NewCloudSQLInstance to instantiate.
+type CloudSQLInst struct {
```
enocom, line 29: This type exists to make it easier to configure mock responses, right? If that's the case, I'd like to name it to make it clear it's a testing convenience, and not related at all to the production code. Maybe just `FakeInstance` or `TestInstance`?

kurtisvg: It seems like it's already clear from `mock.CloudSQLInstance`, but okay.

internal/mock/sqladmin.go
kurtisvg: It's protecting the whole struct, which is why I embedded it vs just over certain fields.

internal/mock/cloudsql.go
kurtisvg: It doesn't matter yet, it almost certainly will at some point in the (near) future, since certain behavior is database specific (like unix socket path). I started fleshing out full options for customizing this, but stopped since I didn't actually need anything for the existing tests.

enocom (approved):

PR #23, opened 2021-05-12
chore: refactor performRefresh to use a cfg struct

kurtisvg:


internal/cloudsql/instance.go
```diff
@@ -111,16 +111,22 @@ func (r *refreshResult) IsValid() bool {
 	}
 }
 
+type refreshCfg struct {
+	refreshTimeout time.Duration
+
+	clientLimiter *rate.Limiter
```
enocom, line 117: It seems a bit weird to have a client and a limiter on a config type. Also, there's the stutter with `refreshCfg` and `refreshTimeout`. So if we look at this type just by what it contains (a refresh timeout, a clientLimiter, and a client), I think we might have more of a "refresher" type here.

There are a few nice benefits to changing the name to focus on what it does:
1. We could test this type and the refresh logic separately
2. The function signature of `performRefresh` becomes simpler and goes from:

``` go
func performRefresh(ctx context.Context, cn connName, k *rsa.PrivateKey, cfg refreshCfg) (metadata, *tls.Config, time.Time, error) {
// ...
}
```

to:

``` go
func (r *refresher) refresh(ctx context.Context, k *rsa.PrivateKey) (metadata, *tls.Config, time.Time, error) {
// assuming connName is stored as a field on refresher
}
```

3. and we could still save it on `Instance` to access whatever fields we might need.

In short, I think the gist of the refactor here is correct, but the naming needs some work and might reveal a cleaner separation of concerns between the `Instance` type and whatever we want to call what is encapsulated here.

kurtisvg: Ok, so I did this. My big concern is that I'm less confident in how closures handle references in this context. 

Specifically is:
```go
i := 0
foo := func() {
   do(i)
}
```

Resolved identically as:
```go
obj := stuct{
  i: 0
}
foo := func() {
   obj.do()
}
```"

Is `obj.do()` considered a reference to do that exercises `do`, or is that create a function to `obj.do`? If I reassign `refresher` after this closure is created but before it's triggered, does it use the value of obj at creation or at execution?

File: internal/cloudsql/instance.go [hunk 2]
```diff
@@ -185,11 +184,10 @@ func (i *Instance) scheduleRefresh(d time.Duration) *refreshResult {
 	res := &refreshResult{}
 	res.ready = make(chan struct{})
 	res.timer = time.AfterFunc(d, func() {
-		ctx, cancel := context.WithTimeout(context.Background(), i.refreshTimeout)
-		res.md, res.tlsCfg, res.expiry, res.err = performRefresh(ctx, i.client, i.clientLimiter, i.connName, i.key)
-		cancel()
-
+		ctx := context.Background() // TODO: store this in Instance
```
enocom, line 187: Re: the TODO. Or Better pass it through from the client call site so clients could cancel the refresh.

enocom (approved):

internal/cloudsql/instance.go
kurtisvg: Feel free to weigh in here: https://github.com/kurtisvg/cloud-sql-go-connector/issues/5#issuecomment-834820781

kurtisvg: (I'm leaning towards opt3 right now)

PR #27, opened 2021-06-21
chore: add tests using mock API

enocom:
Also, add an example .envrc so it's easy to run tests in this project.

This is just a first set of tests for the outermost interface. I'll follow up with more tests of the instance and refresher types, updating this PR.


internal/mock/cloudsql.go
```diff
@@ -30,45 +35,43 @@ type FakeCSQLInstance struct {
...
+		return nil, nil, tls.Certificate{}, err
 	}
 
 	cert := &x509.Certificate{
 		SerialNumber: &big.Int{},
 		Subject: pkix.Name{
-			Country:      []string{"US"},
-			Organization: []string{"Google, Inc"},
-			CommonName:   "Google Cloud SQL Server CA",
+			CommonName: fmt.Sprintf("%s:%s", project, name),
```
enocom, line 74: Is this right? I thought the region would be included in the common name.

dialer_test.go
```diff
@@ -0,0 +1,155 @@
...
+
+	d, err := NewDialer(context.Background(), WithDefaultDialOptions(WithPublicIP()))
+	if err != nil {
+		t.Fatalf("expected NewDialer to succeed, but got error: %v", err)
+	}
+	replaceSQLAdminService(d, mc, url)
+
+	// instance name is bad
+	_, err = d.Dial(context.Background(), "bad-instance-name")
+	if !errorContains(err, "invalid instance") {
```
enocom, line 97: Checking string values on an error isn't the best. We do want to ensure that we're getting a particular error here. As an alternative, we should probably introduce concrete error types and return those, but that is a bigger change that I wanted to hold off on for now. What do you think?

dialer_test.go
```diff
@@ -0,0 +1,155 @@
...
+// replaceSQLAdminClient patches the Dialer's SQL Admin API client with one that
+// uses the provided HTTP client and endpoint.
+func replaceSQLAdminService(d *Dialer, client *http.Client, endpoint string) {
+	svc, err := sqladmin.NewService(
+		context.Background(),
+		option.WithHTTPClient(client),
+		option.WithEndpoint(endpoint),
+	)
+	if err != nil {
+		panic(err)
```
kurtisvg, line 24: Panic's seem bad - even in tests? Is there a specific reason to use it here?

dialer_test.go
```diff
@@ -0,0 +1,155 @@
...
+		if err := cleanup(); err != nil {
+			t.Fatalf("%v", err)
+		}
+	}()
+
+	d, err := NewDialer(context.Background(), WithDefaultDialOptions(WithPublicIP()))
+	if err != nil {
+		t.Fatalf("expected NewDialer to succeed, but got error: %v", err)
+	}
+	replaceSQLAdminService(d, mc, url)
```
kurtisvg, line 50: I'd rather have this be more explicit that's happening - it seems like `d.sqladmin := SQLAdminService...` is more clear that the dialer is being mutated.

dialer_test.go
```diff
@@ -0,0 +1,155 @@
+package cloudsqlconn
```
kurtisvg, line 1: Need a license on this file

dialer_test.go
```diff
@@ -0,0 +1,155 @@
...
+	"testing"
+
+	"cloud.google.com/cloudsqlconn/internal/mock"
+	"google.golang.org/api/option"
+	sqladmin "google.golang.org/api/sqladmin/v1beta4"
+)
+
+// replaceSQLAdminClient patches the Dialer's SQL Admin API client with one that
+// uses the provided HTTP client and endpoint.
+func replaceSQLAdminService(d *Dialer, client *http.Client, endpoint string) {
```
kurtisvg, line 17: Is this just to save some boilerplate for converting `mc, url` into a service? If so, we should consider having a function that just skips a step and returns the SQLAdminService

dialer_test.go
kurtisvg: +1 to more concrete error types.

dialer_test.go
```diff
@@ -0,0 +1,155 @@
...
+func TestDialerInstantiationErrors(t *testing.T) {
+	_, err := NewDialer(context.Background(), WithCredentialsFile("bogus-file.json"))
+	if err == nil {
+		t.Fatalf("expected NewDialer to return error, but got none.")
+	}
+}
+
+func errorContains(err error, want string) bool {
+	return strings.Contains(err.Error(), want)
+}
```
kurtisvg, line 76: This function seems borderline unnecessary.

dialer_test.go
```diff
@@ -0,0 +1,155 @@
...
+
+func TestDialWithAdminAPIErrors(t *testing.T) {
+	inst := mock.NewFakeCSQLInstance("my-project", "my-region", "my-instance")
+	mc, url, cleanup := mock.HTTPClient()
+	stop := mock.StartServerProxy(mock.ServerProxyConfig{
+		Instance: inst,
+	})
+	defer func() {
+		stop()
+		_ = cleanup() // skip mock verification of HTTP methods
```
kurtisvg, line 86: We shouldn't skip this here, because if it throws an error it means an unexpected call was made.

dialer_test.go
```diff
@@ -0,0 +1,155 @@
...
+	if !errorContains(err, "invalid instance") {
+		t.Fatalf("expected Dial to fail with bad instance name, but it succeeded.")
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel()
+
+	// context is canceled
+	_, err = d.Dial(ctx, "my-project:my-region:my-instance")
+	if !errorContains(err, "context canceled") {
```
kurtisvg, line 106: I think a canceled context has a specific error type?

dialer_test.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	stop := mock.StartServerProxy(mock.ServerProxyConfig{
+		Instance:    inst,
+		InvalidCert: true,
+	})
+	defer func() {
+		stop()
+		_ = cleanup() // skip mock verification of HTTP methods
+	}()
```
kurtisvg, line 148: These need to be two different defers - the cleanup should happen even if it exist earlier

internal/mock/cloudsql.go
```diff
@@ -30,45 +35,43 @@ type FakeCSQLInstance struct {
...
 	cert    *x509.Certificate
+	tlsCert tls.Certificate
 }
 
 // NewFakeCSQLInstance returns a CloudSQLInst object for configuring mocks.
-func NewFakeCSQLInstance(project, region, name string) (FakeCSQLInstance, error) {
+func NewFakeCSQLInstance(project, region, name string) FakeCSQLInstance {
 	// TODO: consider options for this?
-	privKey, cert, err := generateInstanceCerts()
+	key, cert, tlsCert, err := generateCerts(project, name)
```
kurtisvg, line 47: The differences between cert and tlscert are not clear to me

internal/mock/cloudsql.go
kurtisvg: It wouldn't surprise me if it's left off considering region isn't used to identify the instances

internal/mock/cloudsql.go
```diff
@@ -77,6 +80,72 @@ func generateInstanceCerts() (*rsa.PrivateKey, *x509.Certificate, error) {
...
+		Bytes: x509.MarshalPKCS1PrivateKey(key),
+	})
+
+	serverCert, err := tls.X509KeyPair(caPEM.Bytes(), caKeyPEM.Bytes())
+	if err != nil {
+		return nil, nil, tls.Certificate{}, err
+	}
 
-	return privKey, cert, nil
+	return key, cert, serverCert, nil
```
kurtisvg, line 103: Again, it's not clear to me why we have cert and serverCert.

Do we still need cert if we are generating the PEMs here? Should we just return the PEM format and use it for signing?

internal/mock/cloudsql.go
```diff
@@ -77,6 +80,72 @@ func generateInstanceCerts() (*rsa.PrivateKey, *x509.Certificate, error) {
...
+}
+
+// StartServerProxy starts a fake server proxy and listens on the provided port
+// on all interfaces, configured with TLS as specified by the FakeCSQLInstance.
+// Callers should invoke the returned function to clean up all resources.
+func StartServerProxy(config ServerProxyConfig) func() {
+	cert := config.Instance.tlsCert
+	if config.InvalidCert {
+		cert = tls.Certificate{}
+	}
```
kurtisvg, line 124: What's the point of this logic? 

Can we just update the instance to have the config instead of relying on a config here?

internal/mock/cloudsql.go
```diff
@@ -77,6 +80,72 @@ func generateInstanceCerts() (*rsa.PrivateKey, *x509.Certificate, error) {
...
+	ln, err := tls.Listen("tcp", ":3307", &tls.Config{
+		Certificates: []tls.Certificate{cert},
+	})
+	if err != nil {
+		panic(err)
```
kurtisvg, line 129: same as above

internal/mock/cloudsql.go
```diff
@@ -77,6 +80,72 @@ func generateInstanceCerts() (*rsa.PrivateKey, *x509.Certificate, error) {
...
+	ctx, cancel := context.WithCancel(context.Background())
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			default:
+				conn, err := ln.Accept()
+				if err != nil {
+					return
```
kurtisvg, line 140: Can we log this somewhere? Maybe the fakeinstance server takes in a `t *testing` as an arg?

internal/mock/sqladmin.go
```diff
@@ -118,7 +118,7 @@ func InstanceGetSuccess(i FakeCSQLInstance, ct int) *Request {
 	db := &sqladmin.DatabaseInstance{
 		BackendType:     "SECOND_GEN",
 		ConnectionName:  fmt.Sprintf("%s:%s:%s", i.project, i.region, i.name),
-		DatabaseVersion: i.dbVersion,
+		DatabaseVersion: i.version,
```
kurtisvg, line 121: `db` is a useful qualifier on version for me

kurtisvg (changes requested):

internal/mock/cloudsql.go
enocom: I confirmed with openssl against my own database and see that it's correct.

dialer_test.go
enocom: If you can stand the string checking for now, I'll follow up with a PR just with errors after we get the testing work complete.

dialer_test.go
enocom: For test setup functions, especially when the chance of failure is small, using panic avoids the boilerplate of checking for errors and then failing the test. We could ignore the error, but in the very small chance something has gone wrong, a panic stops the tests from resulting in hard to understand errors. For these kind of functions (e.g., generating an RSA key), we don't really care to know what the error is, just that it didn't happen.

dialer_test.go
enocom: I agree. That reads better.

dialer_test.go
enocom: Yes, as written. I'll expand this to check that err isn't nil in a subsequent commit.

And if you can hold your nose for a few commits, we'll remove this entirely in favor of concrete types.

dialer_test.go
enocom: Turns out, I had mistakenly registered mocks in a previous attempt. I'll add the check back.

dialer_test.go
enocom: Yes. Fixed.

dialer_test.go
enocom: Fixed.

internal/mock/cloudsql.go
enocom: As in pass the instance in as a parameter here? I think I like that. It does mean we'll have to export the fields on the instance.

internal/mock/sqladmin.go
enocom: :+1:

internal/mock/cloudsql.go
enocom: In a later commit, I end up making Cert and Key public, which means we can build a TLS config in the mock listener when we need it. So, I'll drop the tlsCert here and build it when it's needed.

dialer_test.go
enocom: If you can bear with me for another PR or two, I'd like to show you how this pattern plays out. If you still don't like it on the last PR, I'll update with `t` pointers and an associated `Fatalf`.

enocom:
@kurtisvg How should we handle the missing application default credentials? We could put some stub credentials into a `testdata` folder and put at it with an environment variable.

enocom:
Added a stub set of credentials in https://github.com/kurtisvg/cloud-sql-go-connector/pull/27/commits/14a7e2c4c6b1cb76bb1fa040943a6fa8e1e3dc61. Once we have a proper CI set up, this commit can be reverted.

kurtisvg:
> @kurtisvg How should we handle the missing application default credentials? We could put some stub credentials into a `testdata` folder and put at it with an environment variable.

Can we give it an empty oauth2 token source or something? We don't actually check the token in the fake API, so it doesn't need to be a valid one

kurtisvg:
> @kurtisvg How should we handle the missing application default credentials? We could put some stub credentials into a `testdata` folder and put at it with an environment variable.

Can we give it an empty oauth2 token source or something? We don't actually check the token in the fake API, so it doesn't need to be a valid one

enocom:
@kurtisvg This is ready for another round.

dialer_test.go
```diff
@@ -0,0 +1,180 @@
+// Copyright 2020 Google LLC
```
kurtisvg, line 1: New file, new date.
```suggestion
// Copyright 2021 Google LLC
```

dialer_test.go
kurtisvg: Ok

dialer_test.go
```diff
@@ -0,0 +1,180 @@
...
+	_, err = d.Dial(ctx, "my-project:my-region:my-instance")
+	if !errors.Is(err, context.Canceled) {
+		t.Fatalf("expected Dial to fail with canceled context, but it succeeded.")
+	}
+
+	// failed to retrieve metadata or ephemeral cert (not registered in the mock)
+	_, err = d.Dial(context.Background(), "my-project:my-region:my-instance")
+	if !errorContains(err, "fetch metadata failed") {
+		t.Fatalf("expected Dial to fail with missing metadata")
+	}
```
kurtisvg, line 136: nit: Should these tests be table driven?

kurtisvg (approved):

dialer_test.go
enocom: Yes, I think so. Let me get concrete error types in here first, and then I'll turn this into a table-driven test.

PR #28, opened 2021-07-15
chore: export items in internal interface

enocom:
To improve testability, this commit exports a number of things within
the internal interface. All the exported things remain internal and
cannot be used outside the package.

- Export cloudsql.ConnName and use it in place of raw strings
- Hide IP Address type lookup inside cloudsql.Instance
- Export cloudsql.Metadata and its fields
- Export Refresher and r.PerformRefresh
- Rename genVerifyPeerCertificateFunc to verifyPeerCertFunc

All these changes (except the rename) will make it possible to improve
test coverage.

### DIFF
--- a/dialer.go
+++ b/dialer.go
@@ -60,7 +60,7 @@ func getDefaultKeys() (*rsa.PrivateKey, error) {
 // Use NewDialer to initialize a Dialer.
 type Dialer struct {
 	lock           sync.RWMutex
-	instances      map[cloudsql.ConnName]*cloudsql.Instance
+	instances      map[string]*cloudsql.Instance
 	key            *rsa.PrivateKey
 	refreshTimeout time.Duration
 
@@ -107,7 +107,7 @@ func NewDialer(ctx context.Context, opts ...DialerOption) (*Dialer, error) {
 	}
 
 	d := &Dialer{
-		instances:      make(map[cloudsql.ConnName]*cloudsql.Instance),
+		instances:      make(map[string]*cloudsql.Instance),
 		key:            cfg.rsaKey,
 		refreshTimeout: cfg.refreshTimeout,
 		sqladmin:       client,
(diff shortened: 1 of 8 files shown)


enocom:
Taking @kurtisvg off until I add the missing test changes.

enocom:
This is ready for review. I've tried to keep it as small as possible.

There is one PR after this which will add a significant amount of new tests.

internal/cloudsql/refresh.go
```diff
@@ -141,16 +145,16 @@ func createTLSConfig(inst connName, m metadata, cert tls.Certificate) *tls.Confi
...
 }
 
-// genVerifyPeerCertificateFunc creates a VerifyPeerCertificate func that verifies that the peer
-// certificate is in the cert pool. We need to define our own because of our sketchy non-standard
-// CNs.
-func genVerifyPeerCertificateFunc(cn connName, pool *x509.CertPool) func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
-	return func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
+// verifyPeerCertFunc creates a VerifyPeerCertificate func that verifies that the peer
+// certificate is in the cert pool. We need to define our own because CloudSQL
+// instances use self-signed certificates as a RootCA.
```
kurtisvg, line 155: We need the verify function not because it's a self-signed cert, but instead because we use the instance name instead of a valid domain for the certificate name field.

kurtisvg (changes requested):
I don't understand why most of these exports are being made. Perhaps it would be easier to include the changes with an example test as to show how it's improving the tests.

enocom:
It will be easier for both of us to look at these changes in the context of all the back-filled tests. The diff is large, but ultimately comes down to the interface change to ConnectInfo (that I'll fix to avoid the performance penalty you mentioned above) and lots of new tests.

dialer.go
```diff
@@ -109,7 +107,7 @@ func NewDialer(ctx context.Context, opts ...DialerOption) (*Dialer, error) {
 	}
 
 	d := &Dialer{
-		instances:      make(map[string]*cloudsql.Instance),
+		instances:      make(map[cloudsql.ConnName]*cloudsql.Instance),
```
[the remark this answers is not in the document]
enocom: Good point. I'll revert this change.

internal/cloudsql/refresh.go
enocom: Fixed.

enocom:
Closing this in favor of a smaller set of PRs.

PR #29, opened 2021-07-21
chore: simplify ConnectInfo interface

enocom:


internal/cloudsql/instance_test.go
```diff
@@ -127,8 +127,14 @@ func TestRefreshTimeout(t *testing.T) {
 		t.Fatalf("failed to initialize Instance: %v", err)
 	}
 
-	_, _, err = im.ConnectInfo(ctx)
+	_, _, err = im.ConnectInfo(context.Background(), PublicIP)
```
kurtisvg, line 130: nit: this should reuse `ctx` since it's the context for the entire test.

kurtisvg (approved):

PR #30, opened 2021-07-21
feat: use cloud.google.com/go/cloudsqlconn

enocom:


enocom:
I've also added code to enforce the [canonical import path](https://golang.org/doc/go1.4#canonicalimports) using the same pattern in the Go client libraries. See [here](https://github.com/googleapis/google-cloud-go/tree/master/storage) for example.

kurtisvg (approved):

PR #31, opened 2021-07-21
chore: improve instance tests

enocom:
- Verify non-error return values from ConnectInfo
- Generate an RSA key in mock package for faster tests
- Introduce mock.TestClient to reduce test setup boilerplate

### DIFF
--- a/dialer_test.go
+++ b/dialer_test.go
@@ -138,6 +138,7 @@ func TestDialWithAdminAPIErrors(t *testing.T) {
 
 func TestDialWithConfigurationErrors(t *testing.T) {
 	inst := mock.NewFakeCSQLInstance("my-project", "my-region", "my-instance")
+	inst.Cert.NotAfter = time.Now().Add(-time.Hour)
 	mc, url, cleanup := mock.HTTPClient(
 		mock.InstanceGetSuccess(inst, 2),
 		mock.CreateEphemeralSuccess(inst, 2),
@@ -168,7 +169,6 @@ func TestDialWithConfigurationErrors(t *testing.T) {
 		t.Fatalf("expected Dial to fail with connection error")
 	}
 
-	inst.Cert.NotAfter = time.Now().Add(-time.Hour)
 	stop := mock.StartServerProxy(t, inst)
 	defer stop()
 
--- a/internal/cloudsql/instance_test.go
+++ b/internal/cloudsql/instance_test.go
@@ -107,11 +107,7 @@ func TestConnectInfoErrors(t *testing.T) {
 	if err != nil {
 		t.Fatalf("%s", err)
 	}
-	defer func() {
-		if err := cleanup(); err != nil {
-			t.Fatalf("%v", err)
-		}
-	}()
+	defer cleanup()
 
 	// Use a timeout that should fail instantly
 	im, err := NewInstance("my-project:my-region:my-instance", client, mock.RSAKey, 0)
--- a/internal/cloudsql/instance_test.go
+++ b/internal/cloudsql/instance_test.go
@@ -73,7 +73,7 @@ func TestConnectInfo(t *testing.T) {
 
 	i, err := NewInstance("my-project:my-region:my-instance", client, mock.RSAKey, 30*time.Second)
 	if err != nil {
-		t.Fatalf("failed to create instance: %v", err)
+		t.Fatalf("failed to create mock instance: %v", err)
 	}
 
 	gotAddr, gotTLSCfg, err := i.ConnectInfo(ctx, PublicIP)
--- a/internal/cloudsql/instance_test.go
+++ b/internal/cloudsql/instance_test.go
@@ -16,12 +16,26 @@ package cloudsql
 
 import (
 	"context"
+	"crypto/rand"
+	"crypto/rsa"
 	"errors"
 	"testing"
 	"time"
 
 	"cloud.google.com/go/cloudsqlconn/internal/mock"
 )
+
+// genRSAKey generates an RSA key used for test.
+func genRSAKey() *rsa.PrivateKey {
+	key, err := rsa.GenerateKey(rand.Reader, 2048)
+	if err != nil {
+		panic(err) // unexpected, so just panic if it happens
+	}
+	return key
+}
+
+// RSAKey is used for test only.
+var RSAKey = genRSAKey()
 
 func TestParseConnName(t *testing.T) {
 	tests := []struct {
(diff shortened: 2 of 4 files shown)


enocom:
The additions to the sqladmin package will see heavy use in the forthcoming refresher tests.

internal/mock/sqladmin.go
```diff
@@ -148,12 +160,20 @@ func InstanceGetSuccess(i FakeCSQLInstance, ct int) *Request {
...
+// endpoint. It responds with a "StatusOK" and a DatabaseInstance object.
 //
-// https://cloud.google.com/sql/docs/mysql/admin-api/rest/v1beta4/sslCerts/createEphemeral
-func CreateEphemeralSuccess(i FakeCSQLInstance, ct int) *Request {
+// https://cloud.google.com/sql/docs/mysql/admin-api/rest/v1beta4/instances/get
+//
+// To Configure the response from the SQL Admin endpoint, use
+// InstanceGetSuccessWithOpts.
+func InstanceGetSuccess(i FakeCSQLInstance, ct int) *Request {
+	return InstanceGetSuccessWithDatabase(i, ct, &sqladmin.DatabaseInstance{})
```
kurtisvg, line 171: I don't really think relying on the `&sqladmin.DatabaseInstance{}` struct is a good idea - it's a part of an API not under out control, and overlaps with FakeCSQLInstance. I think we should really be treating "FakeCSQLInstance" as  an options holder. If we need to change something in the SQLAdmin.DatabaseInstance{} struct, let's do it by changing FakeCSQLInstance.

internal/mock/sqladmin.go
```diff
@@ -229,3 +249,40 @@ func CreateEphemeralSuccess(i FakeCSQLInstance, ct int) *Request {
...
+	return CreateEphemeralSuccessWithExpiry(i, ct, time.Now().Add(time.Hour))
+}
+
+// TestClient is a convenience wrapper that configures a mock HTTP client and
+// sqladmin.Service based on a connection name.
+func TestClient(proj, reg, name string, db *sqladmin.DatabaseInstance, certExpiry time.Time) (*sqladmin.Service, func() error, error) {
+	inst := NewFakeCSQLInstance(proj, reg, name)
+	mc, url, cleanup := HTTPClient(
+		InstanceGetSuccessWithDatabase(inst, 1, db),
+		CreateEphemeralSuccessWithExpiry(inst, 1, certExpiry),
```
kurtisvg, line 268: I don't really like the way we've pushed this into another function. It makes it less clear from each tests perspective what's being blocked.

internal/mock/sqladmin.go
```diff
@@ -229,3 +249,40 @@ func CreateEphemeralSuccess(i FakeCSQLInstance, ct int) *Request {
...
+func genRSAKey() *rsa.PrivateKey {
+	key, err := rsa.GenerateKey(rand.Reader, 2048)
+	if err != nil {
+		panic(err) // unexpected, so just panic if it happens
+	}
+	return key
+}
+
+// RSAKey is used for test only.
+var RSAKey = genRSAKey()
```
kurtisvg, line 288: nit: In general, I think globals should try to stay towards the top of the file.

That said, this seems like a regression to me - if each instance has it's own key, they intentionally cant be mixed up

kurtisvg (changes requested):

enocom:
To summarize an offline discussion. I'm going to make a few adjustments here:

1. add a new function to make it easy to get a mock sql admin service (leaving HTTPClient because we still need it)
2. use the FakeCSQLInstance instead of the sqladmin.DatabaseInstance
3. introduce functional options to the FakeCSQLInstance intitializer to make it easy to configure various details.

internal/mock/cloudsql.go
```diff
@@ -75,6 +75,7 @@ func NewFakeCSQLInstance(project, region, name string, opts ...FakeCSQLInstanceO
 		project:   project,
 		region:    region,
 		name:      name,
+		ipAddrs:   map[string]string{"PUBLIC": "0.0.0.0"},
```
kurtisvg, line 78: nit: Should we define "PUBLIC" and "PRIVATE" in a constant somewhere?

dialer_test.go
```diff
@@ -138,6 +138,7 @@ func TestDialWithAdminAPIErrors(t *testing.T) {
 
 func TestDialWithConfigurationErrors(t *testing.T) {
 	inst := mock.NewFakeCSQLInstance("my-project", "my-region", "my-instance")
+	inst.Cert.NotAfter = time.Now().Add(-time.Hour)
```
kurtisvg, line 141: nit: Should this be an option on `inst`? `WithCertExpiry` or something?

internal/cloudsql/instance_test.go
```diff
@@ -59,70 +55,59 @@ func TestParseConnName(t *testing.T) {
...
 		if err := cleanup(); err != nil {
 			t.Fatalf("%v", err)
 		}
 	}()
 
-	client, err := sqladmin.NewService(ctx, option.WithHTTPClient(mc), option.WithEndpoint(url))
+	i, err := NewInstance("my-project:my-region:my-instance", client, mock.RSAKey, 30*time.Second)
 	if err != nil {
-		t.Fatalf("client init failed: %s", err)
+		t.Fatalf("failed to create instance: %v", err)
```
kurtisvg, line 76:
```suggestion
		t.Fatalf("failed to create mock instance: %v", err)
```

internal/mock/sqladmin.go
```diff
@@ -29,9 +31,13 @@ import (
 	"sync"
 	"time"
 
+	"google.golang.org/api/option"
 	sqladmin "google.golang.org/api/sqladmin/v1beta4"
 )
 
+// RSAKey is used for test only.
+var RSAKey = genRSAKey()
```
kurtisvg, line 39: nit: does this make the most sense in the mock? Or should it be in the test package?

kurtisvg (approved):
All nits

internal/mock/cloudsql.go
enocom: I went back and forth on using the same constants from the `cloudsql` package. Since it's constrained to this file and not something test callers will use, it's not terrible to use a string.

enocom: This will come up again in my next PR. Let's talk about it again to see if it's working as a string.

dialer_test.go
enocom: I've done exactly that in the next PR.

internal/mock/sqladmin.go
enocom: It looks like we'll only be using this in the `internal/cloudsql` tests, so I'll move this key there and we can always move it to a better location if necessary.

Issue #15, opened 2021-05-03
Add support for metrics / tracing


kurtisvg:
See https://github.com/GoogleCloudPlatform/cloudsql-proxy/issues/169

PR #32, opened 2021-07-23
feat: add initial support for tracing

enocom:
This is what the Dial trace looks like:

image

This is what the Refresh trace looks like:

image

Some of the trace code is borrowed from [here](https://github.com/googleapis/google-cloud-go/blob/master/internal/trace/trace.go). Note: the code there uses context lookups to find the associated span, which has a performance penalty. This code uses a closure so callers don't need to keep track of the span.

Related to #15.

It's probably worth discussing how we want to name our traces, too, because they'll become part of the public API in effect. As a result, I've not used concrete type names for internal tracing, just so we can move the internals around without issue in the future.


dialer.go
```diff
@@ -120,23 +121,34 @@ func NewDialer(ctx context.Context, opts ...DialerOption) (*Dialer, error) {
...
+		endInfo(err)
 		return nil, err
 	}
 	addr, tlsCfg, err := i.ConnectInfo(ctx, cfg.ipType)
 	if err != nil {
+		endInfo(err)
 		return nil, err
 	}
-	addr = net.JoinHostPort(addr, serverProxyPort)
+	endInfo(err)
```
kurtisvg, line 145: nit: do you think this would be more readable in an enclosed function?

internal/trace/trace.go
```diff
@@ -0,0 +1,86 @@
...
+		if err != nil {
+			span.SetStatus(toStatus(err))
+		}
+		span.End()
+	}
+}
+
+// toStatus interrogates an error and converts it to an appropriate
+// OpenCensus status.
+// Note: this function is burrowed from
```
kurtisvg, line 44:
```suggestion
// Note: this function is borrowed from
```

kurtisvg (changes requested):
One thing that aren't clear to me:

Are these exported all the time? What if someone wants to export somewhere specific (aka not Cloud Logging)?

dialer.go
enocom: It would be cleaner for the tracing at least. But the result would be a function that took three arguments (context, ipType, and instance)) and returned four (addr, tlsCfg, i. and err). So in this case, it seems better overall to keep these here.

enocom:
> Are these exported all the time? What if someone wants to export somewhere specific (aka not Cloud Logging)?

The tracing is enabled only if a client has used OpenCensus to register an exporter. With no exporter registered, these traces become no-ops. Also, the client can choose their exporter as the tracing here is agnostic.

kurtisvg:
> > Are these exported all the time? What if someone wants to export somewhere specific (aka not Cloud Logging)?
> 
> The tracing is enabled only if a client has used OpenCensus to register an exporter. With no exporter registered, these traces become no-ops. Also, the client can choose their exporter as the tracing here is agnostic.

Let's include instructions in the README for this.

dialer.go
```diff
@@ -120,23 +121,34 @@ func NewDialer(ctx context.Context, opts ...DialerOption) (*Dialer, error) {
 
 // Dial returns a net.Conn connected to the specified Cloud SQL instance. The instance argument must be the
 // instance's connection name, which is in the format "project-name:region:instance-name".
-func (d *Dialer) Dial(ctx context.Context, instance string, opts ...DialOption) (net.Conn, error) {
+func (d *Dialer) Dial(ctx context.Context, instance string, opts ...DialOption) (conn net.Conn, err error) {
+	var endDial trace.EndSpanFunc
+	ctx, endDial = trace.StartSpan(ctx, "cloud.google.com/go/cloudsqlconn.Dial")
```
kurtisvg, line 126: Sorry one comment I forgot to mention - Can we make these specific to an instance? If I have 5 instances, how do I tell which one this is dialing?

kurtisvg (changes requested):

dialer.go
enocom: Definitely. I'll add a tag with the instance. We can annotate the trace with additional information as we like, too.

enocom:
I've added the instance name as an attribute to the parent span in both cases:

image

kurtisvg (approved):

PR #33, opened 2021-08-02
Increase coverage of refresh logic

### LINKED ISSUE #9: Add better tests and improve Code Coverage

enocom:
- Introduce an initializer for refresher
- Add functional options to configure a FakeCSQLInstance:
    - WithPrivateIP
    - WithCertExpiry
    - WithRegion
    - WithFirstGenBackend
    - WithSigner
    - WithMissingIPAddrs

Fixes #9.

### DIFF
--- a/internal/cloudsql/refresh.go
+++ b/internal/cloudsql/refresh.go
@@ -110,7 +110,7 @@ func fetchEphemeralCert(ctx context.Context, client *sqladmin.Service, inst conn
 	}
 	resp, err := client.SslCerts.CreateEphemeral(inst.project, inst.name, &req).Context(ctx).Do()
 	if err != nil {
-		return tls.Certificate{}, fmt.Errorf("create failed: %w", err)
+		return tls.Certificate{}, fmt.Errorf("create ephemeral cert failed: %w", err)
 	}
 
 	// parse the client cert
--- a/internal/cloudsql/refresh.go
+++ b/internal/cloudsql/refresh.go
@@ -110,7 +110,7 @@ func fetchEphemeralCert(ctx context.Context, client *sqladmin.Service, inst conn
 	}
 	resp, err := client.SslCerts.CreateEphemeral(inst.project, inst.name, &req).Context(ctx).Do()
 	if err != nil {
-		return tls.Certificate{}, fmt.Errorf("create ephemeral failed: %w", err)
+		return tls.Certificate{}, fmt.Errorf("create failed: %w", err)
 	}
 
 	// parse the client cert
@@ -154,10 +154,11 @@ func createTLSConfig(inst connName, m metadata, cert tls.Certificate) *tls.Confi
 }
 
 // genVerifyPeerCertificateFunc creates a VerifyPeerCertificate func that verifies that the peer
-// certificate is in the cert pool. We need to define our own because of our sketchy non-standard
-// CNs.
-func genVerifyPeerCertificateFunc(cn connName, pool *x509.CertPool) func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
-	return func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
+// certificate is in the cert pool. We need to define our own because CloudSQL
+// instances use the instance name (e.g., my-project:my-instance) instead of a
+// valid domain name for the certificate's Common Name.
+func genVerifyPeerCertificateFunc(cn connName, pool *x509.CertPool) func(rawCerts [][]byte, _ [][]*x509.Certificate) error {
+	return func(rawCerts [][]byte, _ [][]*x509.Certificate) error {
 		if len(rawCerts) == 0 {
 			return fmt.Errorf("no certificate to verify")
 		}
@@ -180,6 +181,17 @@ func genVerifyPeerCertificateFunc(cn connName, pool *x509.CertPool) func(rawCert
 	}
 }
 
+// newRefresher creates a Refresher.
+func newRefresher(timeout time.Duration, interval time.Duration, burst int, svc *sqladmin.Service) refresher {
+	return refresher{
+		timeout:       timeout,
+		clientLimiter: rate.NewLimiter(rate.Every(interval), burst),
+		client:        svc,
+	}
+}
+
+// refresher manages the SQL Admin API access to instance metadata and to
+// ephemeral certificates.
 type refresher struct {
 	// timeout is the maximum amount of time a refresh operation should be allowed to take.
 	timeout time.Duration
--- a/internal/cloudsql/refresh_test.go
+++ b/internal/cloudsql/refresh_test.go
@@ -173,7 +173,7 @@ func TestRefreshWithFailedMetadataCall(t *testing.T) {
 				mock.NewFakeCSQLInstance(
 					cn.project, cn.region, cn.name,
 					mock.WithRegion("my-region"),
-					mock.WithSigner(func(_ *x509.Certificate, _ *rsa.PrivateKey) ([]byte, error) {
+					mock.WithCertSigner(func(_ *x509.Certificate, _ *rsa.PrivateKey) ([]byte, error) {
 						return nil, nil
 					}),
 				), 1),
@@ -185,7 +185,7 @@ func TestRefreshWithFailedMetadataCall(t *testing.T) {
 				mock.NewFakeCSQLInstance(
 					cn.project, cn.region, cn.name,
 					mock.WithRegion("my-region"),
-					mock.WithSigner(func(_ *x509.Certificate, _ *rsa.PrivateKey) ([]byte, error) {
+					mock.WithCertSigner(func(_ *x509.Certificate, _ *rsa.PrivateKey) ([]byte, error) {
 						certPEM := &bytes.Buffer{}
 						pem.Encode(certPEM, &pem.Block{
 							Type:  "CERTIFICATE",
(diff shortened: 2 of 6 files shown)


enocom:
Before all these test PRs:

- `cloud.google.com/cloudsqlconn`   coverage: **54.8%**
- `cloud.google.com/cloudsqlconn/internal/cloudsql` coverage: **73.6%**

With this PR:

- `cloud.google.com/go/cloudsqlconn` coverage: **82.5%** of statements
- `cloud.google.com/go/cloudsqlconn/internal/cloudsql` coverage: **89.9%** of statements

internal/cloudsql/refresh_test.go
```diff
@@ -15,77 +15,334 @@
...
+	testCases := []struct {
+		req     *mock.Request
+		wantErr string
+		desc    string
+	}{
+		{
+			req:     mock.InstanceGetSuccess(inst, 1), // not an ephemeral cert call
+			wantErr: "create failed",
+			desc:    "When the CreateEphemeralCert call fails",
+		},
```
enocom, line 237: There are three additional test cases to add here, but they'll require registering a separate instance with the mock server such that the metadata call succeeds but the fetch cert call fails (i.e., 1. when MarshalPKIXPublicKey fails, 2. When the cert response fails to decode, and 3. When the X.509 cert does not parse).

internal/cloudsql/refresh.go
```diff
@@ -110,7 +110,7 @@ func fetchEphemeralCert(ctx context.Context, client *sqladmin.Service, inst conn
 	}
 	resp, err := client.SslCerts.CreateEphemeral(inst.project, inst.name, &req).Context(ctx).Do()
 	if err != nil {
-		return tls.Certificate{}, fmt.Errorf("create ephemeral failed: %w", err)
+		return tls.Certificate{}, fmt.Errorf("create failed: %w", err)
```
kurtisvg, line 113: this seems way more vague - create what?

internal/cloudsql/refresh.go
```diff
@@ -180,6 +181,17 @@ func genVerifyPeerCertificateFunc(cn connName, pool *x509.CertPool) func(rawCert
...
 }
 
+// newRefresher creates a Refresher.
+func newRefresher(timeout time.Duration, interval time.Duration, burst int, svc *sqladmin.Service) refresher {
+	return refresher{
+		timeout:       timeout,
+		clientLimiter: rate.NewLimiter(rate.Every(interval), burst),
+		client:        svc,
+	}
+}
```
kurtisvg, line 191: Why are we using an initializer here?

internal/cloudsql/refresh_test.go
```diff
@@ -15,77 +15,334 @@
...
+		{
+			req: mock.InstanceGetSuccess(
+				mock.NewFakeCSQLInstance(
+					cn.project, cn.region, cn.name,
+					mock.WithRegion("my-region"),
+					mock.WithMissingIPAddrs(),
+				), 1),
+			wantErr: "no supported IP addresses",
+			desc:    "When the instance has no supported IP addresses",
```
kurtisvg, line 169: Should this be no supported? Or should it be none of the type specified?

internal/mock/cloudsql.go
```diff
@@ -63,6 +69,54 @@ func WithPublicIP(addr string) FakeCSQLInstanceOption {
...
+// result should be PEM-encoded.
+type SignFunc = func(*x509.Certificate, *rsa.PrivateKey) ([]byte, error)
+
+// WithSigner configures the signing function used to generate a signed
+// certificate.
+func WithSigner(s SignFunc) FakeCSQLInstanceOption {
+	return func(f *FakeCSQLInstance) {
+		f.sign = s
+	}
+}
```
kurtisvg, line 110: nit: This seems like it would be more understandable as "WithCertSigner"

internal/mock/cloudsql.go
```diff
@@ -72,20 +126,56 @@ func NewFakeCSQLInstance(project, region, name string, opts ...FakeCSQLInstanceO
...
-		dbVersion: "POSTGRES_12", // default of no particular importance
-		Key:       key,
-		Cert:      cert,
+		project:     project,
+		region:      region,
+		name:        name,
+		ipAddrs:     map[string]string{"PUBLIC": "0.0.0.0"},
+		dbVersion:   "POSTGRES_12", // default of no particular importance
+		backendType: "SECOND_GEN",
+		sign:        SelfSign,
```
kurtisvg, line 135: nit: Field seems like it should be "signer" to be consistent with the option "WithSigner"

internal/mock/cloudsql.go
```diff
@@ -63,6 +69,54 @@ func WithPublicIP(addr string) FakeCSQLInstanceOption {
...
+		f.region = region
+	}
+}
+
+// WithFirstGenBackend sets the server backend type to FIRST_GEN.
+func WithFirstGenBackend() FakeCSQLInstanceOption {
+	return func(f *FakeCSQLInstance) {
+		f.backendType = "FIRST_GEN"
+	}
+}
```
kurtisvg, line 98: nit: I don't know if we really care about this, there are no first gen instances any more

internal/cloudsql/refresh_test.go
```diff
@@ -15,77 +15,334 @@
...
+		},
+	}
+	for i, tc := range testCases {
+		t.Run(tc.desc, func(t *testing.T) {
+			mc, url, cleanup := mock.HTTPClient(tc.req)
+			client, err := sqladmin.NewService(
+				context.Background(),
+				option.WithHTTPClient(mc),
+				option.WithEndpoint(url),
+			)
```
kurtisvg, line 208: Should this use `NewSQLAdminService`? Should we get rid of NewService since we aren't using it anymore?

internal/cloudsql/refresh_test.go
kurtisvg: Can you change req to a slice of requests instead and include those tests here?

kurtisvg (changes requested):

internal/cloudsql/refresh.go
enocom: This was a dev artifact. I've improved the error message.

internal/cloudsql/refresh.go
enocom: Two reasons: I'm don't want to type all these fields out in test, and it seems better to hide the rate limiter inside the refresher.

internal/mock/cloudsql.go
enocom: Agreed.

internal/mock/cloudsql.go
enocom: Done.

internal/mock/cloudsql.go
enocom: Would you be comfortable removing these lines then?

https://github.com/kurtisvg/cloud-sql-go-connector/blob/cab162906d2bec0862d86a76af1bff2d42d3b2f3/internal/cloudsql/refresh.go#L59-L61

internal/cloudsql/refresh_test.go
enocom: Yes. Done.

internal/cloudsql/refresh_test.go
enocom: I have this in here to cover these lines:

https://github.com/kurtisvg/cloud-sql-go-connector/blob/cab162906d2bec0862d86a76af1bff2d42d3b2f3/internal/cloudsql/refresh.go#L73-L75

We could add a function option to configure a custom IP type, e.g., `mock.WithIPType("foo")`. Perhaps this would be useful for future work? WDYT?

internal/cloudsql/refresh_test.go
enocom: Will do.

enocom:
With the additional tests, here's where we stand relative to coverage now:


| package  | before | after |
|---|---|---|
|cloud.google.com/cloudsqlconn   | 54.8%  | 82.5%  |
|cloud.google.com/cloudsqlconn/internal/cloudsql | 73.6% | 91.1%  |

internal/cloudsql/refresh_test.go
kurtisvg: Ah OK - this seems fine.

internal/cloudsql/refresh_test.go
```diff
@@ -15,77 +15,363 @@
...
+			desc:    "When parsing the cert fails",
+		},
+	}
+	for i, tc := range testCases {
+		mc, url, cleanup := mock.HTTPClient(tc.reqs...)
+		client, err := sqladmin.NewService(
+			context.Background(),
+			option.WithHTTPClient(mc),
+			option.WithEndpoint(url),
+		)
```
kurtisvg, line 274: Should this use `NewSQLAdminService`? 

We may want to just not export HTTPClient anymore, to make sure we are using this new method.

internal/cloudsql/refresh.go
kurtisvg: I'll downgrade this to a nit, but I think it's usage from `instance.go` is harder to interpret with the constructor.

kurtisvg (approved):

internal/cloudsql/refresh_test.go
enocom: Yes. I'll make a pass on all the tests and make sure we're just using `NewSQLAdminService` everywhere.

internal/cloudsql/refresh.go
enocom: I agree the initializer isn't super easy to read. Let's let this sit for awhile and revisit it when the refresher changes. It's possible it's better to just manually configure the refresher in test rather than introduce an initializer.

PR #34, opened 2021-08-03
feat: Add Close method to Dialer

### LINKED ISSUE #5: Add a way to close dialers and instance refreshs
Right now, once an instance is dialer, it will continue to refresh those certificates indefinitely. We should expose some way to close the DialManager in a way that propagates down to the instance objects and stops refreshes:

https://github.com/kurtisvg/cloud-sql-go-connector/blob/7e8955216cc91999e3d8d17ed9eced8f63564ca7/internal/cloudsql/instance.go#L100-L114

kurtisvg:
Fixes #5

### DIFF
--- a/internal/cloudsql/instance.go
+++ b/internal/cloudsql/instance.go
@@ -129,8 +129,8 @@ type Instance struct {
 
 	// ctx is the default ctx for refresh operations. Canceling it prevents new refresh
 	// operations from being triggered.
-	ctx      context.Context
-	closeCtx func()
+	ctx    context.Context
+	cancel context.CancelFunc
 }
 
 // NewInstance initializes a new Instance given an instance connection name
@@ -139,7 +139,7 @@ func NewInstance(instance string, client *sqladmin.Service, key *rsa.PrivateKey,
 	if err != nil {
 		return nil, err
 	}
-	ctx, close := context.WithCancel(context.Background())
+	ctx, cancel := context.WithCancel(context.Background())
 	i := &Instance{
 		connName: cn,
 		key:      key,
@@ -148,8 +148,8 @@ func NewInstance(instance string, client *sqladmin.Service, key *rsa.PrivateKey,
 			clientLimiter: rate.NewLimiter(rate.Every(30*time.Second), 2),
 			client:        client,
 		},
-		ctx:      ctx,
-		closeCtx: close,
+		ctx:    ctx,
+		cancel: cancel,
 	}
 	// For the initial refresh operation, set cur = next so that connection requests block
 	// until the first refresh is complete.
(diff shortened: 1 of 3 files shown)


internal/cloudsql/instance.go
```diff
@@ -127,7 +127,10 @@ type Instance struct {
 	// replacement to occur.
 	next *refreshResult
 
-	// TODO: add a way to close
+	// ctx is the default ctx for refresh operations. Canceling it prevents new refresh
+	// operations from being triggered.
+	ctx      context.Context
+	closeCtx func()
```
enocom, line 133: nit: `cancel` matches the signature `context.WithCancel`.

internal/cloudsql/instance.go
```diff
@@ -136,6 +139,7 @@ func NewInstance(instance string, client *sqladmin.Service, key *rsa.PrivateKey,
 	if err != nil {
 		return nil, err
 	}
+	ctx, close := context.WithCancel(context.Background())
```
enocom, line 142: ditto above: s/close/cancel/

internal/cloudsql/instance_test.go
```diff
@@ -137,3 +137,31 @@ func TestConnectInfoErrors(t *testing.T) {
...
+	defer cleanup()
+
+	// Step 0: Generate Keys
+	key, err := rsa.GenerateKey(rand.Reader, 2048)
+	if err != nil {
+		t.Fatalf("failed to generate keys: %v", err)
+	}
+
+	// Set up an instance and then close it immediately
+	im, err := NewInstance("my-proj:my-region:my-inst", client, key, 30)
```
enocom, line 157: key can be RSAKey here so avoid the cost of generating keys above.

internal/cloudsql/instance.go
```diff
@@ -190,7 +202,7 @@ func (i *Instance) scheduleRefresh(d time.Duration) *refreshResult {
 	res := &refreshResult{}
 	res.ready = make(chan struct{})
 	res.timer = time.AfterFunc(d, func() {
-		ctx := context.Background() // TODO: store this in Instance
+		ctx := i.ctx
```
enocom, line 205: Should we check if this context is canceled here to avoid scheduling another refresh operation otherwise?

enocom: Likewise, should we be stopping the current and next refresh results as well?

kurtisvg: I added a check to prevent the new refresh from being scheduled. It didn't necessarily seem worth it to try to stop the AfterFunc from triggering when scheduled, since the context is closed it'll be a no-op.

enocom (approved):

PR #41, opened 2021-08-13
fix: add support for closing the default dialer

### LINKED ISSUE #39: docs: include note about leaked goroutines in cloudsqlconn.Dial
As a result, the goroutines started by the dialer are leaked.

Possible solution: add a `Close` function that closes the default dialer.

enocom:
Fixes #39.

### DIFF
--- a/dial.go
+++ b/dial.go
@@ -36,6 +36,15 @@ func Dial(ctx context.Context, instance string) (net.Conn, error) {
 	return d.Dial(ctx, instance)
 }
 
+func Close() error {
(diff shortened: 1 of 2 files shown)


kurtisvg (changes requested):
I would like to discuss if this change is necessary. 

My concern is that since you can't re-initialize the default dialer, this is a permanent closure and might cause more problems than it solves.

enocom:
Per our discussion, I'm going to close this in favor a PR with documentation describing the leaked goroutines.

PR #44, opened 2021-08-28
feat: add support for IAM DB Authn

enocom:


.envrc.example
```diff
@@ -2,3 +2,4 @@ export POSTGRES_USER=some-user
 export POSTGRES_PASS=some-password
 export POSTGRES_DB=some-db-name
 export POSTGRES_CONNECTION_NAME=some-project:some-region:some-instance
+export POSTGRES_IAM_USER=some-iam-user
```
kurtisvg, line 5: nit: but I think we should be consistent with these. In the proxy we use one and java/python the other. So:
* `POSTGRES_IAM_USER` or
* `POSTGRES_USER_IAM`

errtypes/errors.go
```diff
@@ -58,9 +58,9 @@ type RefreshError struct {
 
 func (e *RefreshError) Error() string {
 	if e.Err == nil {
-		return fmt.Sprintf("Server error: %v", e.genericError)
+		return fmt.Sprintf("Refresh error: %v", e.genericError)
```
kurtisvg, line 61: These are changes that should be in a separate PR.

errtypes/errors_test.go
```diff
@@ -33,14 +33,14 @@ func TestErrorFormatting(t *testing.T) {
 			want: "Client error: error message (connection name = \"proj:reg:inst\")",
 		},
 		{
-			desc: "server error message without internal error",
+			desc: "refresh error message without internal error",
```
kurtisvg, line 36: These are changes that should be in a separate PR.

dialer_test.go
```diff
@@ -69,6 +69,15 @@ func TestDialerInstantiationErrors(t *testing.T) {
 	if err == nil {
 		t.Fatalf("expected NewDialer to return error, but got none.")
 	}
+	_, err = NewDialer(context.Background(),
```
kurtisvg, line 72: These changes could probably be in a separate PR?

dialer_test.go
```diff
@@ -154,3 +163,44 @@ func TestDialWithConfigurationErrors(t *testing.T) {
...
+  "project_id": "a-project-id",
+  "private_key_id": "a-private-key-id",
+  "private_key": "a-private-key",
+  "client_email": "email@example.com",
+  "client_id": "12345",
+  "auth_uri": "https://accounts.google.com/o/oauth2/auth",
+  "token_uri": "https://oauth2.googleapis.com/token",
+  "auth_provider_x509_cert_url": "https://www.googleapis.com/oauth2/v1/certs",
+  "client_x509_cert_url": "https://www.googleapis.com/robot/v1/metadata/x509/email%40example.com"
+}`)
```
kurtisvg, line 179: nit: This would be more readable if it was declared globally

internal/cloudsql/refresh.go
```diff
@@ -103,10 +105,26 @@ func fetchMetadata(ctx context.Context, client *sqladmin.Service, inst connName)
...
-// fetchEphemeralCert uses the Cloud SQL Admin API's createEphemeral method to create a signed TLS
-// certificate that authorized to connect via the Cloud SQL instance's serverside proxy. The cert
-// if valid for approximately one hour.
-func fetchEphemeralCert(ctx context.Context, client *sqladmin.Service, inst connName, key *rsa.PrivateKey) (c tls.Certificate, err error) {
+func refreshToken(ts oauth2.TokenSource, tok *oauth2.Token) (*oauth2.Token, error) {
+	expiredToken := &oauth2.Token{
+		AccessToken:  tok.AccessToken,
+		TokenType:    tok.TokenType,
+		RefreshToken: tok.RefreshToken,
+		Expiry:       time.Time{}.Add(1), // Expired
```
kurtisvg, line 113: Why do we add one here?

internal/cloudsql/refresh.go
```diff
@@ -118,6 +136,31 @@ func fetchEphemeralCert(ctx context.Context, client *sqladmin.Service, inst conn
...
+		if tokErr != nil {
+			return tls.Certificate{}, errtypes.NewRefreshError(
+				"failed to refresh Oauth2 token",
+				inst.String(),
+				tokErr,
+			)
+		}
+		// TODO: remove this once issue with OAuth2 Tokens is resolved.
+		// See https://github.com/GoogleCloudPlatform/cloudsql-proxy/issues/852.
+		req.AccessToken = strings.TrimRight(tok.AccessToken, ".")
```
kurtisvg, line 162: I _think_ we can leave this out now

internal/cloudsql/refresh_test.go
```diff
@@ -119,6 +121,127 @@ func TestRefreshFailsFast(t *testing.T) {
...
 
+type tokenResp struct {
+	tok *oauth2.Token
+	err error
+}
+
+type fakeTokenSource struct {
+	responses []tokenResp
+	mu        sync.Mutex
+	i         int
```
kurtisvg, line 132: nit: I have no idea what `i` is for if there is also a `count`

options.go
```diff
@@ -42,17 +50,32 @@ func DialerOptions(opts ...DialerOption) DialerOption {
...
-// WithCredentialsFile returns a DialerOption that specifies a service account or refresh token JSON credentials file to be used as the basis for authentication.
+// WithCredentialsFile returns a DialerOption that specifies a service account
+// or refresh token JSON credentials file to be used as the basis for
+// authentication.
 func WithCredentialsFile(filename string) DialerOption {
 	return func(d *dialerConfig) {
-		d.sqladminOpts = append(d.sqladminOpts, apiopt.WithCredentialsFile(filename))
+		b, err := ioutil.ReadFile(filename)
+		if err != nil {
+			d.err = err
```
kurtisvg, line 60: Should this be a config error?

options.go
```diff
@@ -84,6 +109,19 @@ func WithRefreshTimeout(t time.Duration) DialerOption {
 	}
 }
 
+// WithIAMAuthn enables IAM DB Authentication. If no token source has been
+// configured, either using WithTokenSource, WithCredentialsFile, or
+// WithCredentialsJSON, the dialer will use the default token source as defined
+// by https://pkg.go.dev/golang.org/x/oauth2/google#FindDefaultCredentialsWithParams.
```
kurtisvg, line 115: Let's make this slightly less explicit incase we add other future token sources and don't remember to update here. 
```suggestion
// WithIAMAuthn enables automatic IAM Database Authentication. If no token source has been
// configured (such as WithTokenSource, WithCredentialsFile, etc), the dialer will 
// use the default token source as defined
// by https://pkg.go.dev/golang.org/x/oauth2/google#FindDefaultCredentialsWithParams.
```

options.go
```diff
@@ -84,6 +109,19 @@ func WithRefreshTimeout(t time.Duration) DialerOption {
...
+//
+// For documentation on IAM DB Authentication, see
+// https://cloud.google.com/sql/docs/postgres/authentication.
+func WithIAMAuthn() DialerOption {
```
kurtisvg, line 119: nit: Should it be `WithIAMAuthN`?

dialer.go
```diff
@@ -78,6 +80,10 @@ type Dialer struct {
 	// dialerID uniquely identifies a Dialer. Used for monitoring purposes,
 	// *only* when a client has configured OpenCensus exporters.
 	dialerID string
+
+	// tokenSource supplies the OAuth2 token used for IAM DB Authn. If IAM DB
+	// Authn is not enabled, tokenSource is unused.
+	tokenSource oauth2.TokenSource
```
kurtisvg, line 86: This naming seems a little concerning because we're using a mere (non-nil) presence of this value to determine if IAMAuthN is enabled. Can we update the name a bit to better reflect this? Like maybe `iamTokenSource` or `iamDBAuthTs`? (the first one seems better)

kurtisvg (changes requested):

.envrc.example
enocom: Good catch. I'll follow the pattern we already have with `POSTGRES_USER_IAM`.

dialer.go
enocom: Good point. I agree.

dialer_test.go
enocom: I'll send them in a separate PR.

dialer_test.go
enocom: Done.

errtypes/errors.go
enocom: :+1:

internal/cloudsql/refresh.go
enocom: The token validation check within `ReuseTokenSource` assumed a zero value of time is not-expired. https://cs.opensource.google/go/x/oauth2/+/master:token.go;l=127

internal/cloudsql/refresh.go
enocom: I'll remove it.

internal/cloudsql/refresh_test.go
enocom: I'll consolidate to just use one.

options.go
enocom: It probably should be. We won't have the instance connection name in this context, but that's probably OK.

options.go
enocom: Yes.

options.go
enocom: Done.

internal/cloudsql/refresh_test.go
```diff
@@ -119,6 +121,125 @@ func TestRefreshFailsFast(t *testing.T) {
...
+
+	for _, tc := range tcs {
+		t.Run(tc.desc, func(t *testing.T) {
+			ts := &fakeTokenSource{responses: tc.resps}
+			r := newRefresher(time.Hour, 30*time.Second, 1, client, ts)
+			_, _, _, err = r.performRefresh(context.Background(), cn, RSAKey)
+			if err == nil {
+				t.Fatalf("expected get failed error, got = %v", err)
+			}
+			if ts.i != tc.wantCount {
```
kurtisvg, line 236: super-nit: Maybe consider `ts.ct` over `ts.i`

options.go
```diff
@@ -31,6 +36,10 @@ type dialerConfig struct {
 	sqladminOpts   []apiopt.ClientOption
 	dialOpts       []DialOption
 	refreshTimeout time.Duration
+	useIAMAuth     bool
```
kurtisvg, line 39: nit: it's probably more correct to use `useIAMAuthN` here as well, since technically AuthZ is always used

kurtisvg (approved):

dialer.go
```diff
@@ -93,6 +99,23 @@ func NewDialer(ctx context.Context, opts ...DialerOption) (*Dialer, error) {
 	for _, opt := range opts {
 		opt(cfg)
 	}
+	if cfg.err != nil {
+		return nil, cfg.err
+	}
```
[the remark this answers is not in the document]
enocom: That's a good point. Done.

internal/cloudsql/refresh_test.go
enocom: Agreed.

options.go
enocom: Definitely. Done.

PR #58, opened 2021-12-16
fix(trace): use LastValue for open connections

enocom:
This commit replaces an erroneous Sum aggregation for a LastValue
aggregation for open connections.

In addition, this commit encapsulates package-level global state for the
metrics into a MetricsCollector.


enocom:
metrics

Now the number of open conns looks like this (and doesn't require any additional configuration to see correctly).

dialer.go
```diff
@@ -143,13 +151,9 @@ func NewDialer(ctx context.Context, opts ...DialerOption) (*Dialer, error) {
...
-	if err := trace.InitMetrics(); err != nil {
-		// This error means the internal metric configuration is incorrect and
-		// should never be surfaced to callers, as there's nothing actionable
-		// for a caller to do. Ignoring the error seems worse and so we return
-		// it.
-		return nil, err
-	}
+	// The error returned indicates a configuration error inside the library and
+	// will be caught by tests. So we safely ignore it here.
+	mc, _ := trace.NewMetricsCollector()
```
kurtisvg, line 156: I'm not sure it's safe to ignore this error. Assuming the tests catch everything also assumes that the version of the dependency is the same as what we've tested. I don't think that's the case here?

I guess it seems like: if we ignore the error and it happens, it blows up with probably a nil reference? If we don't ignore the error and it happens, they at least get an error with what's wrong?

dialer.go
```diff
@@ -159,7 +163,12 @@ func NewDialer(ctx context.Context, opts ...DialerOption) (*Dialer, error) {
 		dialerID:       uuid.New().String(),
 		iamTokenSource: cfg.tokenSource,
 		dialFunc:       cfg.dialFunc,
+		metrics:        mc,
+		openConns:      make(map[string]int64),
```
kurtisvg, line 167: Would it be better for openConns to be a property on cloudsql.Instance?

dialer.go
```diff
@@ -219,21 +228,25 @@ func (d *Dialer) Dial(ctx context.Context, instance string, opts ...DialOption)
 	}
 	latency := time.Since(startTime).Milliseconds()
 	go func() {
-		trace.RecordDialLatency(ctx, instance, d.dialerID, latency)
-		trace.RecordConnectionOpen(ctx, instance, d.dialerID)
+		d.mu.Lock()
```
kurtisvg, line 231: This puts a mutex in front of every single connection opening or shutting. Can we find something that's per instance instead? Or how about https://pkg.go.dev/sync/atomic?

dialer.go
```diff
@@ -159,7 +163,12 @@ func NewDialer(ctx context.Context, opts ...DialerOption) (*Dialer, error) {
 		dialerID:       uuid.New().String(),
 		iamTokenSource: cfg.tokenSource,
 		dialFunc:       cfg.dialFunc,
+		metrics:        mc,
+		openConns:      make(map[string]int64),
+		done:           make(chan struct{}),
 	}
+	// TODO: don't start the goroutine unless a user has enabled metrics.
+	go d.reportOpenConns(ctx)
```
kurtisvg, line 171: This seems like it'll use the context that constructed the dialer for background work. Are we use we want to do that?

dialer.go
```diff
@@ -255,6 +268,34 @@ func (i *instrumentedConn) Close() error {
 	return nil
 }
 
+func (d *Dialer) reportOpenConns(ctx context.Context) {
+	// TODO: make this metrics reporting interval configurable.
+	t := time.NewTicker(30 * time.Second)
+	defer t.Stop()
+	for range t.C {
```
kurtisvg, line 275: would this make more sense to make for loop unconditional and change the default case to t.stop? Otherwise this goroutine will run for a bit of time after those channels after actually closed.

dialer.go
```diff
@@ -255,6 +268,34 @@ func (i *instrumentedConn) Close() error {
 	return nil
 }
 
+func (d *Dialer) reportOpenConns(ctx context.Context) {
```
kurtisvg, line 271: maybe tangential, but do we need to report this every 30 seconds? Can we just report it when it changes?

kurtisvg (changes requested):

dialer.go
enocom: Yeah, it's cheap to just return it, even if it's unlikely or even impossible.

dialer.go
enocom: This first pass is a terrible idea. I've updated to use atomics which results in fewer lines of code too.

dialer.go
enocom: Done.

dialer.go
enocom: I've dropped the loop in favor or reporting when they metrics happen. This way we let OpenCensus deal with intervals.

dialer.go
enocom: Yes, definitely. I've verified this works just fine too.

dialer.go
enocom: I've just dropped this.

internal/trace/metrics.go
```diff
@@ -14,67 +14,63 @@ var (
...
-	)
-	connectionsView = &view.View{
+	connectionsView := &view.View{
 		Name:        "/cloudsqlconn/open_connections",
-		Measure:     mConnections,
-		Description: "The sum of Cloud SQL connections",
-		Aggregation: view.Sum(),
+		Measure:     mc.mConnections,
+		Description: "The current number of open Cloud SQL connections",
+		Aggregation: view.LastValue(),
```
kurtisvg, line 46: Why are we using LastValue as opposed to Count?

How does this effect things when there are multiple instances of the dialer or the dialer ends unexpectedly. Does it ever return to 0?

kurtisvg (changes requested):

dialer.go
```diff
@@ -219,21 +220,23 @@ func (d *Dialer) Dial(ctx context.Context, instance string, opts ...DialOption)
...
-		trace.RecordConnectionOpen(ctx, instance, d.dialerID)
+		n := atomic.AddUint64(&i.OpenConns, 1)
+		d.metrics.RecordOpenConnections(ctx, int64(n), d.dialerID, i.String())
+		d.metrics.RecordDialLatency(ctx, instance, d.dialerID, latency)
 	}()
 
-	return newInstrumentedConn(tlsConn, instance, d.dialerID), nil
+	return newInstrumentedConn(tlsConn, func() {
+		n := atomic.AddUint64(&i.OpenConns, ^uint64(0))
+		d.metrics.RecordOpenConnections(ctx, int64(n), d.dialerID, i.String())
```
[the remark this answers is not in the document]
enocom: Probably not in case the context was canceled. OpenCensus uses the context to associate traces. Here since we're only concerned with a metric, `context.Background()` is probably acceptable.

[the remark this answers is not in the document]
kurtisvg: aren't we handing `ctx` into `newInstrumentedConn` here though?

internal/trace/metrics.go
enocom: At least for Cloud Monitoring, the LastValue type is the only type that results in a Gauge metric. Everything else, include Count, will be a Cumulative metric.

See https://github.com/census-ecosystem/opencensus-go-exporter-stackdriver/blob/dfce60d8dc4099a38012eb860ffba13f0e52e1a8/stackdriver.go#L33-L34.

I'll test to see what Cloud Monitoring does when a client doesn't report a zero value.

enocom: What happens when a connection stays open for say 8 hours? What happens with the metric?

enocom: When a dialer ends unexpectedly, the gauge metric does not return to zero. I think that's reasonable behavior. For metrics that might have a significant amount of time between them, the [Data Retention policy](https://cloud.google.com/monitoring/quotas#data_retention_policy) for custom metrics is 24 months. So I think the scenario for longer lived connections is well supported.

kurtisvg: Does the datapoint show up if it's outside of the graph window in pantheon? If I set the window to 1h, but the value was lated updated 8 hours previously does it still show up?

kurtisvg (approved):

internal/trace/metrics.go
kurtisvg: @enocom this would be good to know but not necessarily blocking

enocom: So I established two connections and let them sit for eight hours. When I set the Cloud Monitoring window to after the first point, I do see the metric appearing, which implies that the Stackdriver exporter is sending points repeatedly, even though they're not changing. Just when my program was about to close the connection, the TCP socket had been closed, so the value did not return to zero. Nonetheless, I think we're good here.

PR #63, opened 2022-01-26
chore: support -short in tests

enocom:


e2e_postgres_test.go
```diff
@@ -12,9 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//go:build !skip_e2e && !skip_postgres
-// +build !skip_e2e,!skip_postgres
```
kurtisvg, line 16: Seems like we should leave `!skip_postgres`?

kurtisvg (changes requested):

e2e_postgres_test.go
enocom: Totally reasonable. Done.

kurtisvg (approved):

PR #66, opened 2022-02-01
chore: create CODEOWNERS

enocom:


.github/CODEOWNERS
```diff
@@ -0,0 +1 @@
+* @kurtisvg @enocom
```
kurtisvg, line 1: Do we want this to be `infra-db-dpes`?

enocom: Sure. That's a better idea.

kurtisvg (approved):

PR #68, opened 2022-02-01
chore(ci): set up release-please

kurtisvg:


enocom (commented):
I assume when we release it will be v1. Yes?

cf https://go.dev/doc/modules/version-numbers#v0-number

enocom (approved):

PR #70, opened 2022-02-05
feat: add custom driver for MySQL

enocom:


e2e_mysql_test.go
```diff
@@ -0,0 +1,70 @@
...
+		t.Skip("skipping MySQL integration tests")
+	}
+	testConn := func(db *sql.DB) {
+		var now time.Time
+		if err := db.QueryRow("SELECT NOW()").Scan(&now); err != nil {
+			t.Fatalf("QueryRow failed: %v", err)
+		}
+		t.Log(now)
+	}
+	mysql.RegisterDialContext("cloudsql-mysql", cloudsqlconn.Dial)
```
kurtisvg, line 59: Does "RegisterDialContext" work for anything besides this go-sql-driver page? If not, I would recommend "cloudsql" as the mysql part is redundant.

kurtisvg (approved):

mysql/mysql/mysql.go
```diff
@@ -0,0 +1,80 @@
...
+// Open accepts a DSN using the go-sql-driver/mysql format. See
+// https://github.com/go-sql-driver/mysql#dsn-data-source-name for details.
+// Note the protocol should match the name used when registering a driver. For
+// example, a connection string might look like this where "cloudsql-mysql" is
+// the named used when registering the driver:
+//
+//     my-user:mypass@cloudsql-mysql(my-proj:us-central1:my-inst)/my-db
+//
+func (d *mysqlDriver) Open(name string) (driver.Conn, error) {
+	drv := &mysql.MySQLDriver{}
```
enocom, line 61: We manually initialize a driver here to mimic what happens when a user imports the MySQL package (see [here](https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/pull/80)).

enocom: This allows us to encapsulate the MySQL driver under whatever name was registered above.

kurtisvg: IMO, it's better to initialize the driver once rather than create a new one every time we need to call open. 

I would probably move it into the mysqlDriver struct, so it's created when the driver is registered.

kurtisvg (changes requested):

mysql/mysql/mysql.go
enocom: Agreed.

mysql/mysql/mysql.go
```diff
@@ -0,0 +1,77 @@
...
+	"cloud.google.com/go/cloudsqlconn"
+	"github.com/go-sql-driver/mysql"
+)
+
+// RegisterDriver registers a MySQL driver that uses the cloudsqlconn.Dialer
+// configured with the provided options. The choice of name is entirely up to
+// the caller and may be used to distinguish between multiple registrations of
+// differently configured Dialers.
+func RegisterDriver(name string, opts ...cloudsqlconn.Option) error {
+	sql.Register(name, &mysqlDriver{})
```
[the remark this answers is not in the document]
enocom: Woops. That's a mistake.

kurtisvg (approved):

PR #80, opened 2022-02-07
chore: update renovate.confg

enocom:


enocom:
I dropped the constraint that enforces Go 1.16 from here: https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/main/.github/renovate.json#L19.

kurtisvg (approved):

PR #101, opened 2022-02-08
fix!: use singular name for package

enocom:
> In go, package names are not plural. This is surprising to programmers who came from other languages and are retaining an old habit of pluralizing names. Don’t name a package httputils, but httputil!

See https://rakyll.org/style-packages.

### DIFF
--- a/dialer.go
+++ b/dialer.go
@@ -25,7 +25,7 @@ import (
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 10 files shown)


kurtisvg (changes requested):
Looks like you missed some in "dialer_test"

kurtisvg (approved):

PR #120, opened 2022-02-15
docs: add pgxpool example

### LINKED ISSUE #119: Usage with a pgxpool (postgres)
## Question
First, I'm very glad to have found this project. It's great! Really hoping it makes it to official status and part of the GCP CloudSQL docs soon.

I'm using a connection pool in my app, but didn't see any documentation for using the connector with a pool. And while I was able to get things working, seemingly without issue, I wanted to run this past you all to make sure I haven't missed any important details or unintended situations.

As far as I can tell, there's only one small change needed when using pgxpool instead of pgx:

```go
dsn := "...connection string..."

// single config points to a 'pgx.ConnConfig' struct
config, err := pgx.ParseConfig(dsn)

// vs

// pool config points to a 'pgxpool.Config' struct
config, err := pgxpool.ParseConfig(dsn)
```

The `pgxpool.Config` struct doesn't expose the `DialFunc` directly, but instead through an additional pointer to a `pgx.ConnConfig`

So instead of
```go
config.DialFunc = func(ctx context.Context, network string, instance string) (net.Conn, error) {
        // ... cloudsqlconn dialer...
}
```

The pool setup would be:
```go
// access the inner ConnConfig field of the pgxpool.Config struct
config.ConnConfig.DialFunc = func(ctx context.Context, network string, instance string) (net.Conn, error) {
	// ...cloudsqlconn dialer...
}
```
Do you see this as the appropriate usage of a `pgxpool` with the `cloud-sql-go-connector`?

I'm not sure if the `pgxpool.Config.BeforeConnect` func would be a better place to modify the underlying `ConnConfig`. I don't think it would since the `DialFunc` isn't really dynamic; doesn't need to be re-defined/bound once the app has initialized. It appears the `cloudsqlconn` dialer is handling all the IAM cert/token housekeeping anyway...

(My app appears to be functioning as expected so far. Additional details below.)

Thanks for the hard work! It's a life saver

## Additional Context
My full setup is as follows:
* AppEngine Standard
* Go 1.15+
* cloudsqlconn v0.1.0
* CloudSQL for Postgres 14.x
* Private IP (GCP internal) to CloudSQL (via VPC connector)
* Cloud IAM User as Postgres Role making the connection (same service account is also associated with the app engine service itself)
* pgx v4 go module (pgxpool)

#### Dialer setup
(For reference) My dialer is configured as follows:
```go
ctx := context.Background()
d, err = cloudsqlconn.NewDialer(ctx, cloudsqlconn.WithIAMAuthN())
...
// config is *pgxpool.Config
config.ConnConfig.DialFunc = func(ctx context.Context, network string, instance string) (net.Conn, error) {
	return d.Dial(ctx, "...<con string>...", cloudsqlconn.WithPrivateIP())
}
...
// pool created as:
pool, connErr := pgxpool.ConnectConfig(ctx, config)
...
```

enocom:
Fixes #119.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ Option](#using-options) below.
 
 To use the dialer with [pgx](https://github.com/jackc/pgx), use
 [pgxpool](https://pkg.go.dev/github.com/jackc/pgx/v4/pgxpool) by configuring
-a [pgConn.DialFunc field][pgconn-cfg] like so:
+a [Config.DialFunc][dial-func] like so:
 
 ``` go
 // Configure the driver to connect to the database
@@ -82,7 +82,7 @@ if err != nil {
 defer conn.Close()
 ```
 
-[pgconn-cfg]: https://pkg.go.dev/github.com/jackc/pgconn#Config
+[dial-func]: https://pkg.go.dev/github.com/jackc/pgconn#Config
 
 ### MySQL
 
--- a/README.md
+++ b/README.md
@@ -84,6 +84,35 @@ if connErr != nil {
 defer conn.Close(ctx)
 ```
 
+To use the dialer with [pgxpool](https://pkg.go.dev/github.com/jackc/pgx/v4/pgxpool), the configuration is similar:
+
+``` go
+// Configure the driver to connect to the database
+dsn := fmt.Sprintf("user=%s password=%s dbname=%s sslmode=disable", pgUser, pgPass, pgDB)
+config, err := pgxpool.ParseConfig(dsn)
+if err != nil {
+    log.Fatalf("failed to parse pgx config: %v", err)
+}
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -84,6 +84,35 @@ if connErr != nil {
 defer conn.Close(ctx)
 ```
 
+To use the dialer with [pgxpool](https://pkg.go.dev/github.com/jackc/pgx/v4/pgxpool), the configuration is similar:
```
kurtisvg, line 87: Should this be the default example if we want people to use pooling?

kurtisvg (approved):

enocom:
Revised the example and fixed some headings (levels were wrong, inconsistent wording etc).

kurtisvg (approved):

PR #125, opened 2022-02-21
feat!: update version policy to latest 3 versions

kurtisvg:
Update version policy to match [google-cloud-go](https://github.com/googleapis/google-cloud-go#go-versions-supported).


enocom (approved):

enocom:
Should this policy also land in the Proxy?

PR #128, opened 2022-02-24
chore: tests run only on supported Go versions

enocom:


kurtisvg (approved):
This needs to be "feat: drop support and testing for Go 1.14"

PR #133, opened 2022-02-25
Add metrics for refresh success and failure

enocom:


internal/trace/metrics.go
```diff
@@ -43,6 +47,16 @@ var (
 		"A failure to dial a Cloud SQL instance",
 		stats.UnitDimensionless,
 	)
+	mSuccessfulRefresh = stats.Int64(
+		"/cloudsqlconn/admin_api/refresh",
```
kurtisvg, line 51: 2 things:

1. Why admin api?
2. Should this be "refresh_success"? And "A successful refresh operation"?

kurtisvg (changes requested):

internal/trace/metrics.go
enocom: 1. admin api was a speculative attempt at clarifying the meaning of "refresh", but I agree it doesn't add much. So I'll update the metric descriptions and remove this path element.
2.Yes. That makes it more consistent with the failure metric.

internal/trace/metrics.go
```diff
@@ -66,6 +80,20 @@ var (
 		Aggregation: view.Count(),
 		TagKeys:     []tag.Key{keyInstance, keyDialerID},
 	}
+	refreshCountView = &view.View{
+		Name:        "/cloudsqlconn/refresh_count",
```
kurtisvg, line 84: This should be "refresh_success_count"

internal/trace/metrics.go
```diff
@@ -66,6 +80,20 @@ var (
...
+		Measure:     mSuccessfulRefresh,
+		Description: "The number of successfully completed certificate refresh operations",
```
kurtisvg, line 86: super-nit: 
```suggestion
		Description: "The number of successful certificate refresh operations",
```

kurtisvg (approved):

PR #154, opened 2022-03-25
chore: rename refresh result to refresh operation

enocom:


enocom:
In addition to this change, I'd like to consolidate all the values that come from performRefresh into a refreshResult type.

kurtisvg (approved):

PR #164, opened 2022-04-04
fix: prevent unnecessary allocation of conn config

enocom:


enocom:
We missed one unnecessary allocation.

FYI @stepan-romankov

kurtisvg (approved):

PR #169, opened 2022-04-06
chore: fix yaml syntax

enocom:


enocom:
This will fix https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/runs/5842090125?check_suite_focus=true.

PR #226, opened 2022-06-14
fix: support MySQL diver's conn check.

### LINKED ISSUE #225: Add support for checking conn health with MySQL Client
## Bug Description

Connecting to CloudSQL MySQL using the connector does not work with [connCheck()](https://github.com/go-sql-driver/mysql/blob/ad9fa14acdcf7d0533e7fbe58728f3d216213ade/conncheck.go#L22) in [go-sql-driver/mysql](https://github.com/go-sql-driver/mysql). This results in the driver not emitting `driver.ErrBadConn`, leading to no retry on bad connections.

`connCheck()` uses implementation details for unix socket to check if connection is alive by reading 1 byte from the tcp connection. This works when the connection is in plain text with underlying unix socket (`syscall.Conn`). They also have special handling for their built-in tls. 

The connector returns `instrumentedConn` on `Dial`. There is no handling for this type in the driver, so the driver skips the connection check.

## Example code (or command)

```go
db := // connect to mysql db with the connector

db.SetMaxOpenConns(1)
db.SetMaxIdleConns(1)

for i := 0; i < 100; i++ {
	var value string
	err = db.QueryRow("SELECT 'thing' as value").Scan(&value)
	if err != nil {
		panic(err)
	}

	log.Info().Msg("ok")
	time.Sleep(10 * time.Second)
}
```

## Stacktrace
```
11:32AM INF ok
11:32AM INF ok
{"level":"debug","time":"2022-06-13T11:32:24-04:00","caller":"/<path>/vendor/github.com/go-sql-driver/mysql/packets.go:37","message":"unexpected EOF"}
panic: invalid connection
```

## How to reproduce

  1. start the polling script 
  2. kill the sql connection (connection closes on the server side, e.g. `wait_timeout`)
     - `SELECT * FROM INFORMATION_SCHEMA.PROCESSLIST WHERE USER='<the-db-user>';`
     - `KILL <process-id>`
 
## Expected behaviour
(works when running `./cloud_sql_proxy` separately)
```
11:39AM INF ok
{"level":"debug","time":"2022-06-13T11:39:30-04:00","caller":"/<path>/vendor/github.com/go-sql-driver/mysql/packets.go:123","message":"closing bad idle connection: EOF"}
11:39AM INF ok
```

## Workaround

Keep no open idle connection to avoid server side disconnects `db.SetMaxIdleConns(0)`

## Environment

1. OS type and version: macOS 12.4, and build from `golang:1.18-alpine`
2. Go version: `go version go1.18 darwin/amd64`
2. Connector version: `v0.4.0` 
3. `go-sql-driver/mysql` version: `v1.6.0`

enocom:
Fixes #225.

### DIFF
--- a/dialer.go
+++ b/dialer.go
@@ -25,6 +25,7 @@ import (
 	"strings"
 	"sync"
 	"sync/atomic"
+	"syscall"
 	"time"
 
 	"cloud.google.com/go/cloudsqlconn/errtype"
@@ -230,7 +231,7 @@ func (d *Dialer) Dial(ctx context.Context, instance string, opts ...DialOption) 
 		trace.RecordDialLatency(ctx, instance, d.dialerID, latency)
 	}()
 
-	return newInstrumentedConn(tlsConn, func() {
+	return newInstrumentedConn(conn, tlsConn, func() {
 		n := atomic.AddUint64(&i.OpenConns, ^uint64(0))
 		trace.RecordOpenConnections(context.Background(), int64(n), d.dialerID, i.String())
 	}), nil
@@ -264,9 +265,10 @@ func (d *Dialer) Warmup(ctx context.Context, instance string, opts ...DialOption
 
 // newInstrumentedConn initializes an instrumentedConn that on closing will
 // decrement the number of open connects and record the result.
-func newInstrumentedConn(conn net.Conn, closeFunc func()) *instrumentedConn {
+func newInstrumentedConn(rawConn, conn net.Conn, closeFunc func()) *instrumentedConn {
 	return &instrumentedConn{
 		Conn:      conn,
+		rawConn:   rawConn,
 		closeFunc: closeFunc,
 	}
 }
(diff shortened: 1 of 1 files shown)


dialer.go
```diff
@@ -275,9 +277,18 @@ func newInstrumentedConn(conn net.Conn, closeFunc func()) *instrumentedConn {
...
 	net.Conn
+	// rawConn is the underlying net.Conn without TLS
+	rawConn   net.Conn
 	closeFunc func()
 }
 
+// SyscallConn supports a connection check in the MySQL driver by delegating to
+// the underlying non-TLS net.Conn.
+func (i *instrumentedConn) SyscallConn() (syscall.RawConn, error) {
+	sconn := i.rawConn.(syscall.Conn)
```
Gentoli, line 288: will this break on windows?

enocom: Good question. I'll a unit test to make sure we have support in either case.

enocom: Looks like the MySQL driver doesn't have support for Windows here: https://github.com/go-sql-driver/mysql/blob/ad9fa14acdcf7d0533e7fbe58728f3d216213ade/conncheck_dummy.go#L9. What would your expectation be?

enocom: I've updated this so it won't panic if the type assertion fails. Separately, this builds on Windows with the unit test, so I think we're good.

enocom: I'll take another look at this and see what we can do. Thanks for all your help on this.

kurtisvg: It looks like the platforms that don't support the call won't use connCheck: https://github.com/go-sql-driver/mysql/blob/ad9fa14acdcf7d0533e7fbe58728f3d216213ade/conncheck_dummy.go#L16

So it seems like it's probably fine to return an error, since if it is being used something probably wrong?

enocom: That’s my understanding as well. @Gentoli what do you think?

Gentoli: Sorry I did not realize `// +build` is built-in to golang. Both with or without the check on the cast should work.

kurtisvg (approved):

PR #237, opened 2022-06-30
feat: expose the WithQuotaProject dialer option

hessjcg:
This option is available as google.golang.org/api/option to the
gcloud go api. It needs to be exposed in the cloud sql connector
also.

### DIFF
--- a/options.go
+++ b/options.go
@@ -139,6 +139,14 @@ func WithAdminAPIEndpoint(url string) Option {
 	}
 }
 
+// WithQuotaProject allows you to change the gcloud project used to track the
+// api request quota.
+func WithQuotaProject(p string) Option {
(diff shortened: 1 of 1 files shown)


options.go
```diff
@@ -139,6 +139,14 @@ func WithAdminAPIEndpoint(url string) Option {
 	}
 }
 
+// WithQuotaProject allows you to change the gcloud project used to track the
+// api request quota.
```
kurtisvg, line 143:
```suggestion
// WithQuotaProject returns an Option that specifies the project used for quota and billing purposes.
```

kurtisvg (approved):

enocom:
FYI: Go likes to capitalize all its acronyms, so get used to writing API, HTTP, IP, etc. ;-)

PR #252, opened 2022-07-20
chore: don't verify mock in error tests

enocom:
This test failed in a recent build because the mock server doesn't always receive the expected number of requests. Removing verification solves the flaky test build.


internal/mock/sqladmin.go
```diff
@@ -53,8 +53,11 @@ func httpClient(requests ...*Request) (*http.Client, string, func() error) {
 		},
 	))
 	// cleanup stops the test server and checks for uncalled requests
```
kurtisvg, line 55:
```suggestion
	// cleanup stops the test server and optionally checks for uncalled requests
```

internal/cloudsql/instance_test.go
```diff
@@ -156,7 +156,7 @@ func TestConnectInfoErrors(t *testing.T) {
 	if err != nil {
 		t.Fatalf("%s", err)
 	}
-	defer cleanup()
+	defer cleanup(true)
```
kurtisvg, line 159: So taking a second look, it seems like we ignore the error in this situations. In that case, do we really need a true/false option or can we just ignore the received value?

We should probably pick one and be consistent - we either pass in a true/false to ignore uncalled endpoints (maybe preferable on the off chance it gets a different error some day), or just all cleanup and ignore the output.

kurtisvg (changes requested):

internal/cloudsql/instance_test.go
enocom: Sorry -- I was mindlessly typing on the last commit. Yes, let's just cleanup and ignore the error. We'll introduce an argument when we need it, but until this, a simple cleanup without an error check seems best.

dialer_test.go
```diff
@@ -126,9 +126,7 @@ func TestDialWithAdminAPIErrors(t *testing.T) {
 	stop := mock.StartServerProxy(t, inst)
 	defer func() {
 		stop()
-		if err := cleanup(); err != nil {
-			t.Fatalf("%v", err)
-		}
+		cleanup()
```
kurtisvg, line 129: nit: Would it be better to use `_ := cleanup()` to indicate we know if returns an error (we just explicitly don't care)

kurtisvg (approved):

PR #262, opened 2022-07-26
chore: fix flaky test with longer refresh timeout

enocom:


enocom:
Here's the build where we saw this error: https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/runs/7509547579?check_suite_focus=true

Error message was `dialer_test.go:206: when TLS handshake fails, want = *errtype.DialError, got = refresh failed: context deadline exceeded`

Source of error is here: https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/blob/main/internal/cloudsql/refresh.go#L366

kurtisvg (approved):

PR #310, opened 2022-09-14
fix: throw error when Auto IAM AuthN is unsupported

### LINKED ISSUE #283: Connector should error on invalid Auto IAM AuthN invocations
If a caller enables Auto IAM AuthN for an engine that does not support it (e.g., SQL Server), it should return an error immediately.

enocom:
Fixes #283.


dialer.go
```diff
@@ -195,17 +209,20 @@ func (d *Dialer) Dial(ctx context.Context, instance string, opts ...DialOption)
 		endInfo(err)
 		return nil, err
 	}
-	addr, tlsCfg, err := i.ConnectInfo(ctx, cfg.ipType)
+	ci, err := i.ConnectInfo(ctx, cfg.ipType)
```
kurtisvg, line 212: We already have [`i.InstanceVersion`](https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/blob/97e4b6538bf085148c90016cce21876e03be97e5/internal/cloudsql/instance.go#L223) - should we just that instead of mutating this return?

dialer.go
```diff
@@ -195,17 +209,20 @@ func (d *Dialer) Dial(ctx context.Context, instance string, opts ...DialOption)
...
 	if err != nil {
 		endInfo(err)
 		return nil, err
 	}
 	endInfo(err)
+	if cErr := verifyIAMAuthNConfig(ci.Version, cfg.refreshCfg.UseIAMAuthN); cErr != nil {
```
kurtisvg, line 218: Any reason not to inline this check?

dialer.go
```diff
@@ -195,17 +209,20 @@ func (d *Dialer) Dial(ctx context.Context, instance string, opts ...DialOption)
...
+		return nil, errtype.NewConfigError("instance does not support Auto IAM AuthN", instance)
```
kurtisvg, line 219: This should return the instance version (e.g. "MYSQL_7 does not support Auto IAM AuthN".

kurtisvg: (AuthN should probably be spelled out here too I guess)

dialer.go
```diff
@@ -170,6 +170,20 @@ func NewDialer(ctx context.Context, opts ...Option) (*Dialer, error) {
 	return d, nil
 }
 
+func verifyIAMAuthNConfig(version string, wantsIAMAuthN bool) error {
+	// Don't check engine version if IAM AuthN isn't requested.
+	if !wantsIAMAuthN {
+		return nil
+	}
+	v := strings.ToLower(version)
```
kurtisvg, line 178: This is probably unnecessary because all the types are currently capitalized: https://cloud.google.com/sql/docs/mysql/admin-api/rest/v1beta4/SqlDatabaseVersion

so we can just use `strings.HasPrefix(v, "POSTGRES")`. 

If we are going to sanitize I'd rather be to caps since it should be a noop

kurtisvg (changes requested):

dialer.go
enocom: Woops -- missed that.

dialer.go
enocom: My only reason is that Dial is getting long and this function adds just enough branches that pulling it out helps readability.

enocom:
@jackwotherspoon Pointed out that a request to generate an ephemeral cert with an access token will return a 40x response for SQL server. So this check probably needs to move into the refresh logic.

dialer.go
```diff
@@ -170,6 +170,21 @@ func NewDialer(ctx context.Context, opts ...Option) (*Dialer, error) {
 	return d, nil
 }
 
+// verifyIAMAuthNConfig checks that the engine support automatic IAM authn. If
+// auto IAM authn was not request, this is a no-op.
+func verifyIAMAuthNConfig(version string, wantsIAMAuthN bool) error {
+	// Don't check engine version if IAM AuthN isn't requested.
+	if !wantsIAMAuthN {
+		return nil
+	}
```
kurtisvg, line 179: nit: this should be up a level (so you don't need to pass in wantsIAMAuth

Then you can rename the function (supportsAutoIAM) which is a bit more clear

enocom:
I moved the check right next to the metadata call and now we get a good error even if the ephemeral certificate call fails.

kurtisvg (approved):

internal/cloudsql/refresh.go
```diff
@@ -375,3 +382,14 @@ func (r refresher) performRefresh(ctx context.Context, cn connName, k *rsa.Priva
...
 }
+
+// supportsAutoIAMAuthN checks that the engine support automatic IAM authn. If
+// auto IAM authn was not request, this is a no-op.
+func supportsAutoIAMAuthN(version string) error {
+	switch {
+	case strings.HasPrefix(version, "POSTGRES"):
+		return nil
+	default:
+		return fmt.Errorf("%s does not support Auto IAM AuthN", version)
```
kurtisvg, line 393:
```suggestion
		return fmt.Errorf("%s does not support Auto IAM DB Authentication", version)
```

kurtisvg (approved):

PR #311, opened 2022-09-14
chore: fix assertion to avoid pointer to error

enocom:
Calling errors.As with a pointer to an error is a mistake now enforced by the compiler in Go 1.19.


internal/cloudsql/refresh_test.go
```diff
@@ -334,8 +334,17 @@ func TestRefreshWithFailedMetadataCall(t *testing.T) {
...
+				if !errors.As(err, &we) {
+					t.Errorf("[%v] PerformRefresh failed with unexpected error, want = %T, got = %v", i, tc.wantErr, err)
+				}
+			case *errtype.ConfigError:
+				if !errors.As(err, &we) {
+					t.Errorf("[%v] PerformRefresh failed with unexpected error, want = %T, got = %v", i, tc.wantErr, err)
+				}
+			default:
+				t.Fatalf("unexpected error type %T", we)
 			}
```
kurtisvg, line 348: I think we want to use `errors.Is` here.

IIUC, [`errors.As`](https://pkg.go.dev/errors#As) does the following:
> As finds the first error in err's chain that matches target, and if one is found, sets target to that error value and returns true. Otherwise, it returns false.

`errors.Is` does the following:
> Is reports whether any error in err's chain matches target.

So `As` attempts to coerce an error into `target` type while `Is` just checks if the error could be `target` type. That's why As wants a pointer - it's modifying it.

kurtisvg (changes requested):

internal/cloudsql/refresh_test.go
enocom: Since we're instantiating a new error when this code runs, I think we want to stick with `errors.As`. If we had a specific single error value that was always returned (e.g. `io.EOF`), `errors.Is` would be appropriate. Since we don't have a single error, `errors.As` let's us assert on the error's type.

See https://go.dev/play/p/Gxe_RIzx3fj

enocom: I've just split the two tests to avoid the headaches with converting an `error` type into a concrete type.

kurtisvg (approved):

PR #331, opened 2022-09-30
chore: add golangci-lint configuration

enocom:


enocom:
Here's what golangci-lint looks like. I've purposely left the build failing to show what errors appear as.

We'll start by making this an optional build step and after some tuning will make it required.

kurtisvg:
Does this replace our current lint flow? (or should it just be added to it?)

enocom:
The new linter will catch files that not have been gofmt'd with gorimports. See [here](https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/actions/runs/3176323054/jobs/5175485517).

enocom:
This will mostly extend our existing lint setup. Checking `go mod tidy` is still required, but golangci-lint will check goimports for us.

hessjcg (approved):
LGTM

PR #346, opened 2022-10-17
feat: add WithAutoIP option

### LINKED ISSUE #244: Add support for automatic IP address selection
In the Cloud SQL Auth Proxy v1, when a user does not specify public or private, the proxy assumes [both address types are acceptable](https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/ebe31d4777358da73ce7412496cbc9d19bc88ee7/proxy/certs/certs.go#L106-L108) and automatically [selects the first address](https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/ebe31d4777358da73ce7412496cbc9d19bc88ee7/proxy/certs/certs.go#L293-L300) returned by the API.

This is a feature request to add a similar option so users who rely on this behavior can enable it. Let's add an option like `WithAutomaticIPAddressSelection` and make it clear that we discourage this option.

enocom:
This commit adds an option to support a legacy behavior where a public IP is selected if available, and otherwise the private IP is used.

Fixes #244.

### DIFF
--- a/options.go
+++ b/options.go
@@ -242,7 +242,7 @@ func WithPrivateIP() DialOption {
 
 // WithAutoIP returns a DialOption that selects the public IP if available and
 // otherwise falls back to private IP. This option is present for backwards
-// compatability only and in general should not be used.
+// compatibility only and is not recommended for use in production.
 func WithAutoIP() DialOption {
 	return func(cfg *dialCfg) {
 		cfg.ipType = cloudsql.AutoIP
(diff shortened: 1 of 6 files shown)


options.go
```diff
@@ -240,6 +240,15 @@ func WithPrivateIP() DialOption {
 	}
 }
 
+// WithAutoIP returns a DialOption that selects the public IP if available and
+// otherwise falls back to private IP. This option is present for backwards
+// compatability only and in general should not be used.
```
kurtisvg, line 245:
```suggestion
// compatibility only and is not recommended for use in production.
```

kurtisvg (approved):

PR #372, opened 2022-11-07
fix: handle context cancellations during instance refresh

### LINKED ISSUE #370: Context cancellation ignored causing heavy CPU usage
## Bug Description

I believe I found an issue with how the cloud-sql-go-connector instance refresh functionality handles errors, or more specifically, cancelled context errors. 

I have a process that runs as a long-running web server and uses the cloud-sql-go-connector to connect to postgres Cloud SQL instances. I noticed extremely high CPU usage after running a few tests and then deleting the Cloud SQL instance that it was connecting to. After doing some profiling and looking through the cloud sql connector code, I think I have a decent understanding of the issue (please correct anything that is wrong).

The `fetchMetadata` function attempts to `Get` the instance details using the sqladmin library [here](https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/blob/v1.0.1/internal/cloudsql/refresh.go#L56). Which as expected, if the instance has been deleted, returns an error [here](https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/blob/v1.0.1/internal/cloudsql/refresh.go#L58). Tracing that back a few more functions, I believe it is called from `scheduleRefresh` in [this section of code](https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/blob/v1.0.1/internal/cloudsql/instance.go#L289-L300). It appears if there is any error returned from the refresh, it immediately tries to schedule another one, but since the instance is deleted, it will just enter into an endless refresh loop and attempts to hold on to the last known good connection. Should this be able to handle unrecoverable errors (like deleted instances) and return more gracefully? At first I thought this was the primary issue, but to me it raised a larger question: why was the connector still refreshing in the background at all since I closed the dialer and connection (see example code below)?

The dialer looked to be cancelled properly [here](https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/blob/v1.0.1/dialer.go#L323) which propagated to the instance [close](https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/blob/v1.0.1/internal/cloudsql/instance.go#L191) and would cancel the internal context.

However, [this line](https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/blob/v1.0.1/internal/cloudsql/refresh.go#L312) returns an error if the context was cancelled. And [this code](https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/blob/v1.0.1/internal/cloudsql/instance.go#L290) always treats all errors the same (by scheduling another refresh).

I believe this needs to check if the returned error is `context.Cancelled`, and if it is, just return. If this is accurate, I would love to submit a PR!

## Example code (or command)

```go
package main

import (
	"context"
	"fmt"
	"log"
	"net"
	"net/http"
	"time"

	"cloud.google.com/go/cloudsqlconn"
	"github.com/jackc/pgx/v4/pgxpool"
)

const (
	username = "postgres"
	password = "TODO"
	project  = "TODO"
	region   = "TODO" // e.g. us-central1
	instance = "TODO"
)

func main() {
	handler := func(w http.ResponseWriter, req *http.Request) {
		msg := "success"
		if err := selectOne(); err != nil {
			msg = err.Error()
		}
		w.Write([]byte(msg))
	}

	http.HandleFunc("/", handler)
	log.Fatal(http.ListenAndServe(":8080", nil))
}

func selectOne() error {
	ctx, cancel := context.WithTimeout(context.Background(), time.Minute*2)
	defer cancel()

	dsn := fmt.Sprintf("user=%s password=%s dbname=postgres sslmode=disable", username, password)
	config, err := pgxpool.ParseConfig(dsn)
	if err != nil {
		return fmt.Errorf("failed to parse pgx config: %v", err)
	}

	dialer, err := cloudsqlconn.NewDialer(ctx)
	if err != nil {
		return fmt.Errorf("failed to initialize dialer: %v", err)
	}
	defer dialer.Close()

	connectionName := fmt.Sprintf("%s:%s:%s", project, region, instance)

	config.ConnConfig.DialFunc = func(ctx context.Context, _ string, instance string) (net.Conn, error) {
		return dialer.Dial(ctx, connectionName)
	}

	conn, err := pgxpool.ConnectConfig(context.Background(), config)
	if err != nil {
		return fmt.Errorf("failed to connect: %w", err)
	}
	defer conn.Close()

	if _, err = conn.Exec(ctx, "select 1"); err != nil {
		return err
	}

	return nil
}
```

## Stacktrace

n/a

## How to reproduce

  1. Create a Cloud SQL postgres instance in the GCP console
  2. Run the above example code `go run main.go` (replacing the `const` values appropriately)
  3. Run `curl -XPOST localhost:8080` to trigger the db connection
  4. Delete the Cloud SQL instance in the GCP console
  5. Wait some amount of time (when the ephemeral cert expires, which I believe is 1 hour) and monitor CPU usage of the program

## Environment

1. OS type and version: `macOS 13.0` (this same issue occurs in a linux container deployed to a GKE cluster)
2. Go version: `go version go1.19.2 darwin/amd64`
3. Connector version: `cloud.google.com/go/cloudsqlconn v1.0.1`

jault3:
## Change Description

Context cancellation was being checked after the immediate retries for errored instance refreshes. This moves the check up a bit further so the background refresh process exits once the context is cancelled.

## Checklist

- [x] Make sure to open an issue as a 
  [bug/issue](https://github.com/GoogleCloudPlatform//issues/new/choose) 
  before writing your code!  That way we can discuss the change, evaluate 
  designs, and agree on the general idea.
- [ ] Ensure the tests and linter pass
- [ ] Appropriate documentation is updated (if necessary)

## Relevant issues:

- Fixes #370

### DIFF
--- a/internal/cloudsql/instance.go
+++ b/internal/cloudsql/instance.go
@@ -293,9 +293,17 @@ func (i *Instance) scheduleRefresh(d time.Duration) *refreshOperation {
 		res.md, res.tlsCfg, res.expiry, res.err = i.r.performRefresh(i.ctx, i.connName, i.key, i.RefreshCfg.UseIAMAuthN)
 		close(res.ready)
 
+		select {
+		case <-i.ctx.Done():
+			// instance has been closed, don't schedule anything
+			return
+		default:
+		}
+
 		// Once the refresh is complete, update "current" with working result and schedule a new refresh
 		i.resultGuard.Lock()
 		defer i.resultGuard.Unlock()
+
 		// if failed, scheduled the next refresh immediately
 		if res.err != nil {
 			i.next = i.scheduleRefresh(0)
@@ -308,14 +316,9 @@ func (i *Instance) scheduleRefresh(d time.Duration) *refreshOperation {
 			}
 			return
 		}
+
 		// Update the current results, and schedule the next refresh in the future
 		i.cur = res
-		select {
-		case <-i.ctx.Done():
-			// instance has been closed, don't schedule anything
-			return
-		default:
-		}
 		t := refreshDuration(time.Now(), i.cur.expiry)
 		i.next = i.scheduleRefresh(t)
 	})
--- a/internal/cloudsql/instance_test.go
+++ b/internal/cloudsql/instance_test.go
@@ -303,3 +303,32 @@ func TestRefreshDuration(t *testing.T) {
 		})
 	}
 }
+
+func TestContextCancelled(t *testing.T) {
+	ctx := context.Background()
+
+	client, cleanup, err := mock.NewSQLAdminService(ctx)
+	if err != nil {
+		t.Fatalf("%s", err)
+	}
+	defer cleanup()
+
+	// Set up an instance and then close it immediately
+	im, err := NewInstance("my-proj:my-region:my-inst", client, RSAKey, 30, nil, "", RefreshCfg{})
+	if err != nil {
+		t.Fatalf("failed to initialize Instance: %v", err)
+	}
+	im.Close()
+
+	// grab the current value of next before scheduling another refresh
+	next := im.next
+
+	op := im.scheduleRefresh(time.Nanosecond)
+	<-op.ready
+
+	// if scheduleRefresh returns without scheduling another one,
+	// i.next should be untouched and remain the same pointer value
+	if im.next != next {
+		t.Fatalf("refresh did not return after a closed context. next pointer changed: want = %p, got = %p", next, im.next)
+	}
+}


internal/cloudsql/instance.go
```diff
@@ -293,9 +293,21 @@ func (i *Instance) scheduleRefresh(d time.Duration) *refreshOperation {
...
+		default:
+		}
+
 		// Once the refresh is complete, update "current" with working result and schedule a new refresh
 		i.resultGuard.Lock()
 		defer i.resultGuard.Unlock()
+
+		// schedule the next refresh in the future
+		t := refreshDuration(time.Now(), i.cur.expiry)
+		i.next = i.scheduleRefresh(t)
```
kurtisvg, line 309: I think this refresh needs to stay put, otherwise there will be 2 refreshes scheduled in the error case (that next clause has a return that prevents these statements from being reached)

kurtisvg (changes requested):

internal/cloudsql/instance.go
jault3: Ah, yep thanks for pointing that out. Updated

enocom:
@jault3 If sign the CLA, we can get to this quickly.

jault3:
> @jault3 If sign the CLA, we can get to this quickly.

👍 apologies for the delay. Trying to track this down on my side. I believe my employer has something set up that I need added to. Will post back once I have it sorted out.

enocom:
You'll have to update the commit to include whatever email you used to sign the CLA, I believe. I forced a re-scan and it's still not happy.

jault3:
Do the commits have to use the email address of the `Authorized Contributors Group`? Because I am a member of that group I submitted in that field. And the email I used in my commits (and shown in the CLA output) is the correct one. But its just my email address, not the group address.

enocom:
I'll check my end to see what we can do.

jault3:
@enocom it is now signed!

enocom (approved):

kurtisvg (approved):

PR #392, opened 2022-11-23
fix: revert "fix: support MySQL driver’s conn check. (#226)"

enocom:
This reverts commit 4b48e3bfe7a5bd8c398592f21eb25ac43644e123.

This fix was incorrect. The connection to the Cloud SQL instance is a TLS connection. A recent version of Go made it possible to access the underlying net.Conn, but noted: "writing to or reading from this connection directly will corrupt the TLS session." [1] The change here made it possible for the MySQL driver to inadvertenly read from the underlying connection. [2] Instead, we should be attempting to read from the TLS connection without corrupting it.

[1] https://pkg.go.dev/crypto/tls#Conn.NetConn.
[2] https://github.com/go-sql-driver/mysql/blob/ad9fa14acdcf7d0533e7fbe58728f3d216213ade/conncheck.go#L34

### DIFF
--- a/dialer.go
+++ b/dialer.go
@@ -22,13 +22,11 @@ import (
 	"crypto/rsa"
 	"crypto/tls"
 	_ "embed"
-	"errors"
 	"fmt"
 	"net"
 	"strings"
 	"sync"
 	"sync/atomic"
-	"syscall"
 	"time"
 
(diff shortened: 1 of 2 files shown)


enocom:
False alarm. See #225 for additional details.

PR #394, opened 2022-11-23
feat: add support for pgxv5

enocom:


enocom:
Depends on https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/pull/392.

enocom:
pgv5 doesn't support Go 1.16 and 1.17.

enocom:
One option here would be to introduce build-constraints so at least clients on newer versions of Go could use pgxv5.

enocom:
We can't conditionally specify a dependency though, so this still won't work.

enocom:
https://github.com/jackc/pgx/issues/1413

PR #427, opened 2023-01-06
fix: use handshake context when possible

### LINKED ISSUE #409: Use HandshakeContext when possible
## Feature Description

When performing an explicit TLS handshake, use the [HandshakeContext](https://pkg.go.dev/crypto/tls#Conn.HandshakeContext) method introduced in Go 1.17 instead of [Handshake](https://pkg.go.dev/crypto/tls#Conn.Handshake). The difference is that `HandshakeContext` will respect context deadlines or cancelation.

In a previous iteration of this connector (cloud-sql-proxy), we an issue where Handshake could hang for minutes (like five minutes): https://github.com/GoogleCloudPlatform/cloud-sql-proxy/issues/986.

## Alternatives Considered
I am not able to come up with a workaround as the standard library method `Handshake` is not cancelable (so the goroutine is hung no matter what we do).

## Additional Context
Equivalent fix in cloud-sql-proxy: https://github.com/GoogleCloudPlatform/cloud-sql-proxy/pull/987
Handshake line in this repo: https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/blob/2a88f2399cb569b04a2e898afae82b79b984cdee/dialer.go#L231

enocom:
Fixes #409.

### DIFF
--- a/connect_tls_117.go
+++ b/connect_tls_117.go
@@ -1,4 +1,4 @@
-// Copyright 2022 Google LLC
+// Copyright 2023 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


hessjcg (approved):
LGTM

PR #448, opened 2023-02-03
fix: improve reliability of certificate refresh

jackwotherspoon:
Updating refresh buffer to 4 minutes.

### DIFF
--- a/internal/cloudsql/instance.go
+++ b/internal/cloudsql/instance.go
@@ -27,6 +27,10 @@ import (
 	sqladmin "google.golang.org/api/sqladmin/v1beta4"
 )
 
+// the refresh buffer is the amount of time before a refresh's result expires
+// that a new refresh operation begins.
+const refreshBuffer = 4 * time.Minute
+
 var (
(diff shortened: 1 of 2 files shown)


internal/cloudsql/instance.go
```diff
@@ -115,8 +115,8 @@ type RefreshCfg struct {
 }
 
 // Instance manages the information used to connect to the Cloud SQL instance by periodically calling
-// the Cloud SQL Admin API. It automatically refreshes the required information approximately 5 minutes
-// before the previous certificate expires (every 55 minutes).
+// the Cloud SQL Admin API. It automatically refreshes the required information approximately 4 minutes
+// before the previous certificate expires (every ~56 minutes).
```
jackwotherspoon, line 119: @enocom On second glance, is this sentence still valid? With the current refresh logic it isn't scheduling a refresh every ~55 minutes. It will schedule multiple refreshes throughout the error with the `d/2` logic right?

enocom (changes requested):

enocom (approved):

PR #453, opened 2023-02-08
fix: error when dialer is misconfigured with token source

enocom:
When a dialer is configured with a token source and with Auto IAM AuthN, the dialer will now return an error and require callers to configure the WithIAMAuthNTokenSources option instead.

This change will prevent OAuth2 tokens with the sqladmin scope from being included in the ephemeral certificate.

### DIFF
--- a/dialer.go
+++ b/dialer.go
@@ -21,6 +21,7 @@ import (
 	"crypto/rand"
 	"crypto/rsa"
 	_ "embed"
+	"errors"
 	"fmt"
 	"net"
 	"strings"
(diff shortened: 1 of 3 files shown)


dialer.go
```diff
@@ -97,6 +98,11 @@ type Dialer struct {
 	iamTokenSource oauth2.TokenSource
 }
 
+var (
+	errUseTokenSource    = errors.New("use WithTokenSource when IAM AuthN is not enabled")
+	errUseIAMTokenSource = errors.New("use WithIAMAuthNTokenSources instead of WithTokenSource be used when IAM AuthN is enabled")
```
jackwotherspoon, line 103: just curious as to why `WithTokenSource` is singular while ``WithIAMAuthNTokenSources` is plural? Is there distinct reasoning for this?

enocom: WithTokenSource expects a token with SQL Admin scope. WithIAMAuthNTokenSources expects two tokens: one with SQL Admin scope for Admin API operations and one with SQL Service Login scope for logging into the database as an IAM user. We require two separate token sources to avoid embedding an OAuth2 token with admins scope into the ephemeral certificate just in case in the very unlikely case the client certificate is leak (again extremely unlikely given this is all in memory).

jackwotherspoon: Thanks! That makes sense

jackwotherspoon (approved):

PR #458, opened 2023-02-14
fix: prevent repeated context expired errors

### LINKED ISSUE #384: refresh failed: context deadline exceeded
## Bug Description

As previously stated on https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/issues/381 and https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/issues/370,
we see this error from time to time:

```
rpc error: code = Internal desc = database read error: failed to connect to `host=localhost user=xxx@xxx.iam database=db`: dial error (timeout: refresh failed: context deadline exceeded)
```

And when it happens, the service seems keep being in this "error state" once in a while.

This graph shows how often it could happen for us:
image

## Example code (or command)

How we setup the connection:
```go
        driver := "cloudsql-postgres"

	options := make([]cloudsqlconn.Option, 0)
	options = append(options, cloudsqlconn.WithIAMAuthN())
	options = append(options, cloudsqlconn.WithDefaultDialOptions(cloudsqlconn.WithPrivateIP()))

	cleanup, err := pgxv4.RegisterDriver(driver, options...)
	if err != nil {
		return nil, nil, fmt.Errorf("could not register CloudSQL driver: %w", err)
	}

	dbHost := fmt.Sprintf("%s:%s:%s", cfg.CloudSQL.GCPProject, cfg.CloudSQL.DBRegion, cfg.CloudSQL.InstanceName)
	dbUser := strings.TrimSuffix(cfg.CloudSQL.User, ".gserviceaccount.com")
	connStr := fmt.Sprintf(
		"host=%s user=%s dbname=%s sslmode=disable",
		dbHost,
		dbUser,
		cfg.CloudSQL.DBName,
	)

	db, err := sql.Open(driver, connStr)
	db.SetMaxOpenConns(2)
	db.SetMaxIdleConns(2)
	db.SetConnMaxLifetime(time.Hour)
...
```

## Stacktrace

We don't have strackrace available now but we logged the context deadline before the code calls onto db to do a simple read. This is what we could see:
```
ctx deadline left before calling db: 9962 ms
```
So this means there is around 10s still available but then the call to db failed with 

> failed to connect to `host=localhost user=xxx@xxx.iam database=db`: dial error (timeout: refresh failed: context deadline exceeded)

## How to reproduce
That is difficult to say. Run a long-running service, randomly do simple db read, this shows up from time to time

## Environment

1. OS type and version: google Cloudrun
2. Go version: `go version` v1.19.3
2. Connector version: v1.0.1

---

Thank you, and if you need any other info we could provide, let us know.

enocom:
The Go Connector uses a rate limiter to prevent excessive API usage. When refresh attempts failed a few times, however, the Go Connector would continue to schedule new attempts which would be rate limited. The rate limited error would then result in another refresh attempt being scheduled. As a result, the Go Connector would report numerous "context deadline expired" errors which were in fact rate limiting errors.

This commit moves the rate limiter to the entry point of the refresh code. This change ensures that if there are refresh failures, there are not runaway attempts to refresh again. Instead, when an error occurs, a new refresh attempt will run only once the rate limiter allows it to run.

Fixes #384

### DIFF
--- a/dialer.go
+++ b/dialer.go
@@ -110,7 +110,7 @@ var (
 // RSA keypair is generated will be faster.
 func NewDialer(ctx context.Context, opts ...Option) (*Dialer, error) {
 	cfg := &dialerConfig{
-		refreshTimeout: 30 * time.Second,
+		refreshTimeout: cloudsql.RefreshTimeout,
 		dialFunc:       proxy.Dial,
 		useragents:     []string{userAgent},
 	}
--- a/internal/cloudsql/instance.go
+++ b/internal/cloudsql/instance.go
@@ -38,7 +38,7 @@ const (
 	// enforced by the rate limiter.
 	refreshInterval = 30 * time.Second
 
-	// refreshTimeout is the maximum amount of time to wait for a refresh
+	// RefreshTimeout is the maximum amount of time to wait for a refresh
 	// cycle to complete. This value should be greater than the
 	// refreshInterval.
 	RefreshTimeout = 60 * time.Second
--- a/internal/cloudsql/instance.go
+++ b/internal/cloudsql/instance.go
@@ -34,8 +34,14 @@ const (
 	// expires that a new refresh operation begins.
 	refreshBuffer = 4 * time.Minute
 
-	// refreshInterval is the amount of time between refresh attempts.
+	// refreshInterval is the amount of time between refresh attempts as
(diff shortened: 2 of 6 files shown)


jackwotherspoon (approved):
LGTM, nice fix!

PR #463, opened 2023-02-16
chore: update issue templates

jackwotherspoon:
Updating issue templates to use Github forms.


.github/ISSUE_TEMPLATE/bug-report.yaml
```diff
@@ -0,0 +1,78 @@
...
+      label: Bug Description
+      description: "Please enter a detailed description of the bug, and any information about what behavior you noticed and why it is defective or unintentional."
+    validations:
+      required: true
+  - type: textarea
+    id: example-code
+    attributes:
+      label: Example code (or command)
+      description: "Please paste any useful application code related to the bug below. (if your code is in a public repo, feel free to paste a link!)"
+      render: Go
```
jackwotherspoon, line 44: @enocom I wonder if it is best to have these not render as code but instead be regular markdown so that users can respond with test and code or multiple code blocks? Thoughts?

jackwotherspoon: See `Example Code` section of [Python Connector bug](https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/issues/new?assignees=&labels=type%3A+bug&template=bug-report.yaml&title=Brief+summary+of+what+bug+or+error+was+observed) for context

enocom: Let's use the same approach whatever we do. I'm sure people will misuse it either way.

enocom (approved):

PR #484, opened 2023-03-08
docs: add Postgres codelab example to README

LukeSchlangen:
Added a link to the new Codelab in the ReadMe. I'm not sure if this is the best place for it. Thoughts?


enocom (commented):
Let's condense this and move it to just above "Installation."

We could say, "For a quick example, try out the Go Connector in a Codelab."

Also, let's use [reference links](https://arcticicestudio.github.io/styleguide-markdown/rules/links.html#prefer-reference-links) for the links.

PR #500, opened 2023-03-27
chore: parse connection name before locking + cleanup

enocom:


internal/cloudsql/instance.go
```diff
@@ -359,7 +355,7 @@ func (i *Instance) scheduleRefresh(d time.Duration) *refreshOperation {
 			// used result while it's still valid and potentially
 			// able to provide successful connections. TODO: This
 			// means that errors while the current result is still
-			// valid are surpressed. We should try to surface
+			// valid are surprised. We should try to surface
```
jackwotherspoon, line 358: silly silly
```suggestion
			// valid are suppressed. We should try to surface
```

jackwotherspoon (changes requested):

internal/cloudsql/instance.go
enocom: Fixed.

jackwotherspoon (approved):

PR #522, opened 2023-04-14
chore: performRefresh returns refreshResult

enocom:
This change brings the Go Connector inline with a recent change in the AlloyDB fork.

In addition, this commit standardizes on "refresh operation" and "refresh result."

### DIFF
--- a/internal/cloudsql/instance.go
+++ b/internal/cloudsql/instance.go
@@ -30,8 +30,8 @@ import (
 )
 
 const (
-	// the refresh buffer is the amount of time before a refresh cycle's result
-	// expires that a new refresh operation begins.
+	// the refresh buffer is the amount of time before a refresh cycle's
+	// certificate expires that a new refresh operation begins.
 	refreshBuffer = 4 * time.Minute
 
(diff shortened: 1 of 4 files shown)


internal/cloudsql/refresh.go
```diff
@@ -273,6 +277,15 @@ func newRefresher(
 	}
 }
 
+// refreshResult contains all the resulting data from the refresh cycle.
```
jackwotherspoon, line 280: nit
```suggestion
// refreshResult contains all the resulting data from the refresh operation.
```

internal/cloudsql/instance.go
```diff
@@ -30,8 +30,8 @@ import (
 )
 
 const (
-	// the refresh buffer is the amount of time before a refresh cycle's result
-	// expires that a new refresh operation begins.
+	// the refresh buffer is the amount of time before a refresh cycle's
```
jackwotherspoon, line 33:
```suggestion
	// the refresh buffer is the amount of time before a refresh operation's
```

jackwotherspoon (approved):

PR #565, opened 2023-06-22
feat: add support for PSC connections

jackwotherspoon:
Add a new dialOption for `WithPSC()` to support connections to Cloud SQL using [private service connect](https://cloud.google.com/vpc/docs/private-service-connect).

### DIFF
--- a/internal/cloudsql/refresh_test.go
+++ b/internal/cloudsql/refresh_test.go
@@ -35,6 +35,7 @@ const testDialerID = "some-dialer-id"
 func TestRefresh(t *testing.T) {
 	wantPublicIP := "127.0.0.1"
 	wantPrivateIP := "10.0.0.1"
+	wantPSC := "abcde.12345.us-central1.sql.goog"
 	wantExpiry := time.Now().Add(time.Hour).UTC().Round(time.Second)
 	wantConnName := "my-project:my-region:my-instance"
 	cn, err := ParseConnName(wantConnName)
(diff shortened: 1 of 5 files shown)


jackwotherspoon:
This currently gets the error: 
```
dial tcp: lookup xxxxxxxxxxxx.xxxxxxxxxxxx.us-central1.sql.goog. on 127.0.0.1:53: no such host)
```

Same DNS name is reachable with Cloud SQL Python Connector so definitely is something needed adjustment in Go connector.

internal/cloudsql/refresh_test.go
```diff
@@ -81,6 +83,13 @@ func TestRefresh(t *testing.T) {
 	if wantPrivateIP != gotIP {
 		t.Fatalf("metadata IP mismatch, want = %v, got = %v", wantPrivateIP, gotIP)
 	}
+	gotPSC, ok := rr.ipAddrs[PSC]
+	if !ok {
+		t.Fatalf("metadata IP addresses did not include PSC endpoint")
```
enocom, line 88: t.Fatal

internal/mock/cloudsql.go
```diff
@@ -155,6 +163,7 @@ func NewFakeCSQLInstance(project, region, name string, opts ...FakeCSQLInstanceO
 		region:       region,
 		name:         name,
 		ipAddrs:      map[string]string{"PUBLIC": "0.0.0.0"},
+		DNSName:      "",
```
enocom, line 166: Should we add a good default here?

enocom (approved):
Some minor nits. LGTM.

internal/cloudsql/refresh_test.go
jackwotherspoon: Done

internal/mock/cloudsql.go
enocom: Sounds good to me.

PR #576, opened 2023-07-18
fix: avoid holding lock over IO

enocom:


postgres/pgxv4/postgres.go
```diff
@@ -59,39 +59,38 @@ type pgDriver struct {
...
+		return nil, err
+	}
+	return stdlib.GetDefaultDriver().Open(dbURI)
 
-	p.mu.RLock()
-	dbURI, ok = p.dbURIs[name]
-	p.mu.RUnlock()
+}
 
+func (p *pgDriver) dbURI(name string) (string, error) {
```
jackwotherspoon, line 70: should we adding a comment above this func explaining it? Even if its only a one-line comment

jackwotherspoon (changes requested):

postgres/pgxv4/postgres.go
```diff
@@ -59,39 +59,38 @@ type pgDriver struct {
...
-	config, err := pgx.ParseConfig(name)
-	if err != nil {
-		return nil, err
-	}
-	instConnName := config.Config.Host // Extract instance connection name
-	config.Config.Host = "localhost"   // Replace it with a default value
-	config.DialFunc = func(ctx context.Context, _, _ string) (net.Conn, error) {
-		return p.d.Dial(ctx, instConnName)
-	}
+	if !ok {
```
jackwotherspoon, line 80: should this be an `else` instead of a second `if`?

jackwotherspoon (approved):

Issue #509, opened 2023-04-03
Use exponential backoff in refresh algorithm


enocom:
### Feature Description

On failure, the refresh cycle runs every thirty seconds which can cause cascading failures in high-volume environments when the SQL Admin API quota is breached. This is a request to use exponential backoff with a randomized jitter to prevent thundering herd scenarios.

PR #611, opened 2023-08-29
fix: exponential backoff upon refresh failure (#509)

annafang-google:
Do exponential backoff upon refresh failure. The delay is  2 * previous refresh interval.


hessjcg (changes requested):
It was not in the requirements, but we need a maximum delay probably around 5 minutes. Otherwise, when the service is down the backoff interval could grow to hours or days.

enocom:
Let's do some design work on this across the connectors and revisit. We definitely want to do complete this work though.

annafang-google:
Agreed to you both, Jonathan and Eno. This PoC is premature. Closing it for now.

Issue #639, opened 2023-09-22
Add pgxv5 driver


jackwotherspoon:
We can now showcase pgxv5 since minimum supported Go version of the connector is 1.19 (previously blocked by Go 1.17)

enocom:
@jackwotherspoon We'll want to create a new package pgxv5 since some customers will be pinned on pgxv4.

PR #642, opened 2023-10-03
feat: add pgx v5 support (#639)

annafang-google:


e2e_postgres_test.go
```diff
@@ -162,6 +163,48 @@ func TestPostgresHook(t *testing.T) {
 		}
 		t.Log(now)
 	}
+	pgxv5.RegisterDriver("cloudsql-postgres")
```
enocom, line 166: Could we turn this into a table test?

https://semaphoreci.com/blog/table-driven-unit-tests-go

postgres/pgxv5/postgres_test.go
```diff
@@ -0,0 +1,44 @@
...
+	"database/sql"
+	"log"
+	"time"
+
+	"cloud.google.com/go/cloudsqlconn"
+	"cloud.google.com/go/cloudsqlconn/postgres/pgxv5"
+)
+
+// Example shows how to use cloudsqlpostgres dialer
+func ExamplePostgresConnection() {
```
enocom, line 27: We can omit this change -- it was a mistake to add it for pgxv4.

annafang-google:
The update of the commit takes a long time to be propagated to the PR page. Will close it and reopen.

postgres/pgxv5/postgres_test.go
annafang-google: Removed it for both v4 and v5.

enocom: thank you

e2e_postgres_test.go
```diff
@@ -162,29 +182,71 @@ func TestPostgresHook(t *testing.T) {
...
-	if err != nil {
-		t.Fatalf("sql.Open want err = nil, got = %v", err)
+func TestPostgresV4Hook(t *testing.T) {
+	tests := []struct {
+		driver   string
+		source   string
+		IAMAuthN bool
+	}{
+		{
+			"cloudsql-postgres",
```
enocom, line 209: We should use named properties here:

``` golang
{
    driver: "cloudsql-postgres",
    source: ... // etc
}
```

enocom (approved):
One small change for named properties. I'll approve since it's a minor change.

PR #679, opened 2023-12-04
chore: flatten refresh config into bool

enocom:
The only value that will ever affect a refresh is Auto IAM AuthN. This commit removes the unnecessary struct wrapper in favor of a simple bool or bool pointer.

### DIFF
--- a/dialer.go
+++ b/dialer.go
@@ -343,12 +343,12 @@ func (d *Dialer) instance(cn cloudsql.ConnName, useIAMAuthN *bool) *cloudsql.Ins
 		// Recheck to ensure instance wasn't created or changed between locks
 		i, ok = d.instances[cn]
 		if !ok {
-			var enable bool
+			var useIAMAuthNDial bool
 			if useIAMAuthN != nil {
-				enable = *useIAMAuthN
+				useIAMAuthNDial = *useIAMAuthN
 			}
 			i = cloudsql.NewInstance(cn, d.sqladmin, d.key,
-				d.refreshTimeout, d.iamTokenSource, d.dialerID, enable)
+				d.refreshTimeout, d.iamTokenSource, d.dialerID, useIAMAuthNDial)
 			d.instances[cn] = i
 		} else if useIAMAuthN != nil && *useIAMAuthN != i.UseIAMAuthNDial() {
 			// Update the instance with the new refresh config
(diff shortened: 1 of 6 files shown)


internal/cloudsql/instance.go
```diff
@@ -204,10 +199,13 @@ func (i *Instance) ConnName() ConnName {
 	return i.connName
 }
 
-// RefreshConfig returns the refresh configuration associated with this
-// instance.
-func (i *Instance) RefreshConfig() RefreshConfig {
-	return i.refreshConfig
+// UseIAMAuthNDial reports whether the instance has set IAM AuthN on a per-dial
+// basis. This field can change from one Dial attemtp to another and affects
```
jackwotherspoon, line 203: nit: typo
```suggestion
// basis. This field can change from one Dial attempt to another and affects
```

internal/cloudsql/instance.go
```diff
@@ -262,14 +260,14 @@ func (i *Instance) InstanceEngineVersion(ctx context.Context) (string, error) {
...
-func (i *Instance) UpdateRefresh(c RefreshConfig) {
-	i.refreshLock.Lock()
-	defer i.refreshLock.Unlock()
+func (i *Instance) UpdateRefresh(useIAMAuthNDial bool) {
+	i.mu.Lock()
+	defer i.mu.Unlock()
 	// Cancel any pending refreshes
 	i.cur.cancel()
 	i.next.cancel()
 	// update the refresh config as needed
```
jackwotherspoon, line 269: should we update this from saying "refresh config"?

dialer.go
```diff
@@ -334,27 +332,27 @@ func (d *Dialer) Close() error {
...
-	if !ok || (r != nil && *r != i.RefreshConfig()) {
+	if !ok || (useIAMAuthN != nil && *useIAMAuthN != i.UseIAMAuthNDial()) {
 		d.lock.Lock()
 		// Recheck to ensure instance wasn't created or changed between locks
 		i, ok = d.instances[cn]
 		if !ok {
-			// Create a new instance
-			if r == nil {
-				r = &d.defaultDialConfig.refreshConfig
+			var enable bool
```
jackwotherspoon, line 346: not sure how i feel about this var just being called `enable`?

jackwotherspoon (changes requested):
quick first pass

PR #682, opened 2023-12-07
chore: improve encapsulation of update refresh logic

enocom:
Also, simplify logical branching with a switch statement.

### DIFF
--- a/internal/cloudsql/instance.go
+++ b/internal/cloudsql/instance.go
@@ -258,6 +258,12 @@ func (i *Instance) InstanceEngineVersion(ctx context.Context) (string, error) {
 	return op.result.version, nil
 }
 
+// NeedsUpdateRefresh determines if the existing refresh needs to be updated
+// with a new IAM AuthN value, by either disabling or enabling it.
+func (i *Instance) NeedsUpdateRefresh(useIAMAuthN *bool) bool {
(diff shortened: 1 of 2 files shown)


internal/cloudsql/instance.go
```diff
@@ -258,6 +258,12 @@ func (i *Instance) InstanceEngineVersion(ctx context.Context) (string, error) {
 	return op.result.version, nil
 }
 
+// NeedsUpdateRefresh determines if the existing refresh needs to be updated
+// with a new IAM AuthN value, by either disabling or enabling it.
+func (i *Instance) NeedsUpdateRefresh(useIAMAuthN *bool) bool {
```
jackwotherspoon, line 263: thoughts on naming the `NeedsRefresh`, `NeedsUpdateRefresh` sound a bit confusing to me?

jackwotherspoon: I guess the method is called `UpdateRefresh` so now I see the naming pattern. Still wondering if `NeedsRefresh` is better though, up to you.

jackwotherspoon (approved):

internal/cloudsql/instance.go
enocom: There's a lot more cleanup we can do here. Since this is a private interface, we can explore a better naming scheme once we land the bigger refactor. How does that sound?

PR #686, opened 2023-12-07
fix: ensure cert refresh recovers from sleep

enocom:
When a computer wakes from sleep after more than ~1 hour and a client tries to connect, the internal certificate cache will have an expired certificate. In the past, we erroneously believed that the TLS handshake would fail if the client certificate expired. However, with TLS 1.3, the client is the last to send a handshake message, and so the server is unable to report any errors until the first client read. See https://go.dev/doc/go1.12#tls_1_3.

As a result, clients who tried to connect after sleep would see bad certificate errors which would resolve only once the remaining refresh duration expired (sleep time + remaining duration).

This commit adds code to check if the client certificate is expired and immediately triggers a blocking refresh if it is. This means that in the case where a cached certificate has expired, it will be discarded and a new one will be retrieved before continuing with the connection.

Related to https://github.com/GoogleCloudPlatform/cloud-sql-proxy/issues/1788

### DIFF
--- a/dialer.go
+++ b/dialer.go
@@ -295,6 +295,15 @@ func (d *Dialer) Dial(ctx context.Context, instance string, opts ...DialOption) 
 }
 
 func invalidClientCert(c *tls.Config) bool {
+	// The following conditions should be impossible (no certs, nil leaf), but
+	// just in case there's an unknown edge case, check assumptions before
+	// proceeding.
+	if len(c.Certificates) == 0 {
+		return true
+	}
+	if c.Certificates[0].Leaf == nil {
+		return true
+	}
 	return time.Now().After(c.Certificates[0].Leaf.NotAfter)
 }
 
--- a/dialer.go
+++ b/dialer.go
@@ -231,6 +231,25 @@ func (d *Dialer) Dial(ctx context.Context, instance string, opts ...DialOption) 
 	}
 	endInfo(err)
 
+	// If the client certificate has expired (as when the computer goes to
+	// sleep, and the refresh cycle cannot run), force a refresh immediately.
+	// The TLS handshake will not fail on an expired client certificate. It's
+	// not until the first read where the client cert error will be surfaced.
+	// So check that the certificate is valid before proceeding.
+	if invalidClientCert(tlsConfig) {
+		i.ForceRefresh()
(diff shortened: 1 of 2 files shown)


dialer_test.go
```diff
@@ -550,17 +582,92 @@ func TestDialerRemovesInvalidInstancesFromCache(t *testing.T) {
...
+				tls: &tls.Config{
+					Certificates: []tls.Certificate{{
+						Leaf: &x509.Certificate{
+							// Certificate expired 10 hours ago.
+							NotAfter: time.Now().Add(-10 * time.Hour),
+						},
+					}},
+				},
+			},
+			// Second call errors to validate error path
```
hessjcg, line 612: Would it be clearer if the error path was in a separate test case?

dialer.go
```diff
@@ -275,6 +294,10 @@ func (d *Dialer) Dial(ctx context.Context, instance string, opts ...DialOption)
 	}), nil
 }
 
+func invalidClientCert(c *tls.Config) bool {
+	return time.Now().After(c.Certificates[0].Leaf.NotAfter)
```
hessjcg, line 298: What happens if some part of this statement `c.Certificates[0].Leaf.NotAfter` is nil? Can we guarantee that whenever this function is called, each of these fields will have a value: `c.Certificates`, `c.Certificates[0]`, `c.Certificates[0].Leaf`, `c.Certificates[0].Leaf.NotAfter`?

hessjcg (commented):
Looks good. A couple of questions.

dialer_test.go
enocom: The implementation makes two calls to ConnectInfo. To test this case, we need to ensure the first call returns an expired cert. For the second call, we return an error to test the extra error path as a convenience. Otherwise, we'd have to do a lot more wiring up of TLS config, server TLS, etc, which might be worth doing, but probably not in this PR.

dialer.go
enocom: This same line of code is in v1 proxy, so I think we're safe. However, it's cheap to add those intermediate checks. I'll do that.

enocom: Fixed.

PR #698, opened 2023-12-15
feat: add connection name to public API

### LINKED ISSUE #320: Export instance connection name parsing
The proxy is currently duplicating the parsing log ([link](https://github.com/GoogleCloudPlatform/cloud-sql-proxy/blob/main/internal/proxy/proxy.go#L35-L71)). This is a feature request to make the parsing exported so we can use it in the proxy and avoid the duplication.

enocom:
Fixes #320

### DIFF
--- a/dialer.go
+++ b/dialer.go
@@ -30,6 +30,7 @@ import (
 	"time"
 
 	"cloud.google.com/go/cloudsqlconn/errtype"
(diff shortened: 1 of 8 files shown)


jackwotherspoon (approved):
LGTM

PR #723, opened 2024-02-13
feat: add support for Go 1.22

enocom:


ttosta-google (approved):

jackwotherspoon:
Should we add to description that we are dropping support for Go 1.19? Seems like this PR is doing both...

enocom:
This changes matches our policy: https://github.com/GoogleCloudPlatform/cloud-sql-go-connector?tab=readme-ov-file#supported-go-versions (e.g., last three Go versions).

Since Go maintains a backwards compatibility guarantee, it's probably unnecessary to say much about dropping older versions.

PR #726, opened 2024-02-13
feat: add support for debug logging

enocom:
Add support for optional debug logging. Presently, we will log the following:

## Logging for refresh operations
- When a refresh operation starts
- Before the rate limiter is acquired
- After the rate limiter is acquired
- When a refresh operation finishes
- When a refresh operation errors
- When a refresh operation is canceled
- The ephemeral certificate’s expiration time
- The next scheduled refresh

## Logging for connection attempts
- The current certificate’s expiration
- The result of the invalid certificate check (i.e., is “now” after the expiration)
- The result of connecting to the TCP socket including the IP address / PSC DNS name
- The result of the TLS handshake

## Logging for connection cache
- Adding connection info to the cache
- Removing connection info from the cache


jackwotherspoon (commented):
Should we add a section to README with how to run with debug logging?

jackwotherspoon (approved):
LGTM ✅

PR #732, opened 2024-02-16
feat: add support for TPC

jackwotherspoon:
Add support for configuring TPC universe domain through `WithUniverseDomain` Option. Also added verification to dialer initializer to error if configured universe domain does not match that sourced from credentials.

For TPC workloads running their own service endpoints, one would configure TPC with the connector through `NewDialer` or if using `database/sql` through `RegisterDriver`:

pgx (NewDialer):
```go
 // Create a new dialer with any options
d, err := cloudsqlconn.NewDialer(context.Background(), cloudsqlconn.WithUniverseDomain("my-universe.domain"))
```

database/sql (RegisterDriver):
```go
cleanup, err := mysql.RegisterDriver("cloudsql-mysql", cloudsqlconn.WithUniverseDomain("my-universe.domain"))
```


options.go
```diff
@@ -178,6 +186,16 @@ func WithHTTPClient(client *http.Client) Option {
 func WithAdminAPIEndpoint(url string) Option {
 	return func(d *dialerConfig) {
 		d.sqladminOpts = append(d.sqladminOpts, apiopt.WithEndpoint(url))
+		d.serviceUniverseDomain = ""
```
jackwotherspoon, line 189: I'm not sure that custom endpoints follow any specific pattern do they? If they don't I'm not sure we can parse the universe domain from the URIs in a reliable manner...

jackwotherspoon: In second thought I guess the universe domain would be the suffix in format from the second last `.` to the end of the URI?

`https://sqladmin.googleapis.com` => `googleapis.com`

`https://sqladmin.apis-s3nscloud.io` => `apis-s3nscloud.io`

options.go
```diff
@@ -178,6 +187,16 @@ func WithHTTPClient(client *http.Client) Option {
 func WithAdminAPIEndpoint(url string) Option {
 	return func(d *dialerConfig) {
 		d.sqladminOpts = append(d.sqladminOpts, apiopt.WithEndpoint(url))
+		d.serviceUniverseDomain = strings.SplitN(url, ".", 2)[1]
```
jackwotherspoon, line 190: not sure if this is complex enough logic, just splitting string based on first "." and grabbing suffix as universe domain

options.go
```diff
@@ -178,6 +187,16 @@ func WithHTTPClient(client *http.Client) Option {
...
+	}
+}
+
+// WithUniverseDomain configures the underlying SQL Admin API client to use
+// the provided universe domain.
```
enocom, line 195: Should we add something about Trusted Private Cloud and link to a docs page for more information?

options.go
enocom: We can remove this line like we discussed. You might have to track that this option (or universe) was enabled, though, to error correctly later.

options.go
```diff
@@ -84,6 +87,12 @@ func WithCredentialsJSON(b []byte) Option {
 			d.err = errtype.NewConfigError(err.Error(), "n/a")
 			return
 		}
+		ud, err := c.GetUniverseDomain()
+		if err != nil {
+			d.err = errtype.NewConfigError(err.Error(), "n/a")
+			return
+		}
+		d.authUniverseDomain = ud
```
enocom, line 95: Can we avoid setting this if it's the default universe?

options.go
```diff
@@ -44,6 +44,8 @@ type dialerConfig struct {
 	logger                 debug.Logger
 	iamLoginTokenSource    oauth2.TokenSource
 	useragents             []string
+	authUniverseDomain     string
+	serviceUniverseDomain  string
```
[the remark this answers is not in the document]
enocom: Could we detect the universe from the credentials reliably? If so, maybe we should just assign into one field (if it's not the default universe).

dialer_test.go
```diff
@@ -271,6 +285,65 @@ func TestIAMAuthNErrors(t *testing.T) {
...
+		desc string
+		opts Option
+	}{
+		{
+			desc: "When universe domain matches GDU",
+			opts: WithOptions(WithUniverseDomain("googleapis.com"), WithCredentialsJSON(fakeServiceAccount)),
+		},
+		{
+			desc: "When universe domain from admin api matches GDU",
+			opts: WithOptions(WithAdminAPIEndpoint("https:sqladmin.googleapis.com"), WithCredentialsJSON(fakeServiceAccount)),
```
enocom, line 299: should this be https://

dialer_test.go
```diff
@@ -271,6 +285,65 @@ func TestIAMAuthNErrors(t *testing.T) {
...
+		},
+		{
+			desc: "When TPC universe matches TPC credential domain",
+			opts: WithOptions(WithUniverseDomain("test-universe.test"), WithCredentialsJSON(fakeTPCServiceAccount)),
```
enocom, line 303: Maybe fakeTPCServiceAccount should be a function that takes a domain? That makes the relationship between inputs here more clear.

enocom (changes requested):

options.go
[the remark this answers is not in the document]
jackwotherspoon: but then I'm not sure I would be able to make the error as actionable. We want to tell the user what auth is set to and what the service is set to so they know which one is the problem

dialer_test.go
jackwotherspoon: this has been removed now

options.go
jackwotherspoon: Probably but would that really simplify anything later in the flow? We would still need to set it if it were non default universe.

options.go
jackwotherspoon: added additional details to comment... "Enables Trusted Partner Cloud (TPC)."

dialer_test.go
jackwotherspoon: I might do this as a follow up PR, ideally i should just have `fakeServiceAccount` take in a string for universe domain and then i wouldn't even need `fakeTPCServiceAccount`

dialer_test.go
```diff
@@ -271,6 +285,61 @@ func TestIAMAuthNErrors(t *testing.T) {
...
+}
+
+func TestUniverseDomainErrors(t *testing.T) {
+	tcs := []struct {
+		desc string
+		opts Option
+	}{
+		{
+			desc: "When universe domain does not match GDU credentials",
+			opts: WithOptions(WithUniverseDomain("test-universe.test")),
```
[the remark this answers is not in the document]
enocom: Or, let's update the description to make it clear we're depending on ADC here implicitly.

dialer_test.go
```diff
@@ -271,6 +285,61 @@ func TestIAMAuthNErrors(t *testing.T) {
...
+		},
+		{
+			desc: "When GDU does not match credential domain",
+			opts: WithOptions(WithCredentialsJSON(fakeTPCServiceAccount)),
+		},
+		{
+			desc: "WithUniverseDomain used alongside WithAdminAPIEndpoint",
+			opts: WithOptions(WithUniverseDomain("googleapis.com"), WithAdminAPIEndpoint("https://sqladmin.googleapis.com")),
```
enocom, line 328: Let's wrap long lines to try to keep things at ~80 characters.

options.go
enocom, line 95: Nit: `credentialsUniverse` reads better to me, since "auth" is overloads with Auto IAM AuthN, database authn, and ADC auth.

options.go
enocom: Do the TPC folks have any guidance for parsing the universe from a URL? It would be nice to avoid setting this to empty string, which isn't semantically correct (although I see how you've got it working here).

enocom (changes requested):
A few more minor issues. I see the bulk of this is about bookkeeping so we can report a good error. I would expect the generated client underneath us would handle that. Do you know if that exists or will exist? We might change our approach here depending on the answer.

dialer.go
```diff
@@ -170,6 +176,23 @@ func NewDialer(ctx context.Context, opts ...Option) (*Dialer, error) {
 		cfg.rsaKey = key
 	}
 
+	if cfg.setUniverseDomain && cfg.setAdminAPIEndpoint {
+		return nil, errors.New(
+			"can not use WithAdminAPIEndpoint and WithUniverseDomain Options together, " +
+				"use WithAdminAPIEndpoint (it already contains the universe domain)",
```
jackwotherspoon, line 182: the vscode lint/format did this i guess??

options.go
jackwotherspoon: I assume this will be handled by the underlying APIARY sqladmin library when TPC changes happen in those in the near future

enocom (approved):

PR #760, opened 2024-04-01
refactor: simplify internal interface

enocom:
This is a port of https://github.com/GoogleCloudPlatform/alloydb-go-connector/pull/536


enocom:
Yes good idea. Will do in the commit message.

PR #761, opened 2024-04-01
refactor: move TLS config creation to ConnectionInfo

enocom:
This refactor moves all TLS configuration into the connection info type and leaves room for altering the configuration based on the request connection path, rather than hiding the configuration deep within the code that retrieves the ephemeral certificate.


jackwotherspoon (approved):
LGTM

PR #766, opened 2024-04-08
fix: return a friendly error if the dialer is closed

enocom:
This is a port of https://github.com/GoogleCloudPlatform/alloydb-go-connector/pull/538

### DIFF
--- a/dialer.go
+++ b/dialer.go
@@ -53,6 +53,9 @@ const (
 )
 
 var (
+	// ErrDialerClosed is used when a caller invokes Dial after closing the
+	// Dialer.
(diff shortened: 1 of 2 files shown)


jackwotherspoon (approved):

rosecodym:
This is _extremely_ minor, but did this change cause the doc comment on `Close` to longer be correct? (Specifically, I don't think that "additional dial operations may succeed" anymore.) I only noticed because that comment slowed me down a bit before realizing that we had the same situation as #777.

enocom:
Yep, you're right. We missed that in this PR. Fixed here: https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/pull/785

PR #772, opened 2024-04-15
feat: add support for a lazy refresh

### LINKED ISSUE #770: Add support for a lazy refresh
### Feature Description

When using the Connector in Cloud Run, the Connector's background refresh cannot run reliably when the CPU is throttled.

To resolve this situation, we should support a lazy refresh that retrieves a new certificate if no valid certificate is available and caches the result. No background goroutines should be involved.

### Sample code

_No response_

### Alternatives Considered

_No response_

### Additional Details

_No response_

enocom:
When creating a Dialer with the WithLazyRefresh option, the connection info and ephemeral certificate will be refreshed only when the cache certificate has expired. No background goroutines run with this option, making it ideal for use in Cloud Run and other serverless environments where the CPU may be throttled.

Fixes #770

### DIFF
--- a/options.go
+++ b/options.go
@@ -42,6 +42,7 @@ type dialerConfig struct {
 	refreshTimeout         time.Duration
 	useIAMAuthN            bool
 	logger                 debug.Logger
+	lazyRefresh            bool
 	iamLoginTokenSource    oauth2.TokenSource
 	useragents             []string
 	credentialsUniverse    string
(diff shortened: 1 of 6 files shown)


options.go
```diff
@@ -42,6 +42,7 @@ type dialerConfig struct {
 	refreshTimeout         time.Duration
 	useIAMAuthN            bool
 	logger                 debug.Logger
+	lazyRefresh            bool
```
jackwotherspoon, line 45: do we want this as `lazyRefresh` or `refreshType`? do we see a future where we add another option for refreshes?

jackwotherspoon (approved):

options.go
enocom: As discussed elsewhere, we're keeping this field unexported so we can adapt it in the future if needed.

PR #781, opened 2024-04-29
fix: retry 50x errors with exponential backoff

### LINKED ISSUE #718: "Failed to get instance metadata" error while connection to CloudSQL
### Bug Description

Hi, we're experiencing a weird behavior while trying to connect CloudSQL using your library.

Once in a while, when pods restart, there is an error while trying to connect to the CloudSQL instance:
```
failed to get instance: Refresh error: failed to get instance metadata (connection name = "<some_connection_name>"): Get "https://sqladmin.googleapis.com/sql/v1beta4/projects/<my_project>/instances/<my_instance>/connectSettings?alt=json&prettyPrint=false": compute: Received 500 `Internal Server Error`
```

When these errors happen, they are always preceded by an error in `gke-metadata-server`:
```
[conn-id:<some_id> ip:<some_ip> pod:<some_pod> rpc-id:<some_rpc_id>] "/computeMetadata/v1/instance/service-accounts/default/token?..." HTTP/500: generic::unknown: loading: getFederatedToken("....iam.gserviceaccount.com", "", "", ....): loading: while fetching k8s token: loading: Post "https://<some_ip>/api/v1/namespaces/<namespace>/serviceaccounts/default/token": context deadline exceeded, Reason: "K8S", UserMessage: "Internal Server Error"
```

The problems started to occur last month, and our code that connects to the server has not changed.

Can you please assist us with the issue?

### Example code (or command)

```Go
import (
    "database/sql"

    "cloud.google.com/go/cloudsqlconn"
    "cloud.google.com/go/cloudsqlconn/mysql/mysql"
)

func connect() {
    cleanup, err := mysql.RegisterDriver("cloudsql-mysql", cloudsqlconn.WithCredentialsFile("key.json"))
    if err != nil {
        // ... handle error
    }
    // call cleanup when you're done with the database connection
    defer cleanup()

    db, err := sql.Open(...)
}
```
```


### Stacktrace

_No response_

### Steps to reproduce?

Connect to the CloudSQL instance, as described in the code

### Environment

1. OS type and version: alpine linux 3.17
2. Go version: 1.21
3. Cloud SQL Go Connector version: 1.6.0


### Additional Details

_No response_

enocom:
This commit adds retry behavior to the two SQL Admin API calls. Any response that results in a 50x error will now be retried up to 5 times with exponential backoff and jitter between each attempt.

The formula used to calculate the duration to wait is:

200ms * 1.618^(attempt + jitter)

This calculation matches what the Cloud SQL Proxy v1 does and will not trigger any significant change in load on the backend.

Fixes #718

### DIFF
--- a/internal/cloudsql/refresh.go
+++ b/internal/cloudsql/refresh.go
@@ -62,7 +62,12 @@ func fetchMetadata(
 	var end trace.EndSpanFunc
 	ctx, end = trace.StartSpan(ctx, "cloud.google.com/go/cloudsqlconn/internal.FetchMetadata")
 	defer func() { end(err) }()
-	db, err := client.Connect.Get(inst.Project(), inst.Name()).Context(ctx).Do()
+
+	db, err := retry50x(ctx, func(ctx2 context.Context) (*sqladmin.ConnectSettings, error) {
+		return client.Connect.Get(
+			inst.Project(), inst.Name(),
+		).Context(ctx2).Do()
+	}, exponentialBackoff)
 	if err != nil {
 		return metadata{}, errtype.NewRefreshError("failed to get instance metadata", inst.String(), err)
 	}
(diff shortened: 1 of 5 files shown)


enocom:
For reference, the v1 code is here: https://github.com/GoogleCloudPlatform/cloud-sql-proxy/blob/88c0183c00e4c532243b8a7b4263b6770adcbdda/proxy/certs/certs.go#L155-L194.

jackwotherspoon (approved):
LGTM ✅

PR #806, opened 2024-05-20
fix: remove duplicate refresh operations

### LINKED ISSUE #771: With IAM Authentication enabled, Connector does duplicate refresh
Using the Cloud SQL Proxy with debug logging enabled, you can see that the Go Connector makes two refresh operations for one connection attempt:

```
$ ./cloud-sql-proxy enocom-experiments-304623:us-central1:pg2 --debug-logs --auto-iam-authn
2024/04/15 10:26:49 Authorizing with Application Default Credentials
2024/04/15 10:26:49 [enocom-experiments-304623:us-central1:pg2] Connection info added to cache
2024/04/15 10:26:49 [enocom-experiments-304623:us-central1:pg2] Connection info refresh operation started
2024/04/15 10:26:49 [enocom-experiments-304623:us-central1:pg2] Connection info refresh operation complete
2024/04/15 10:26:49 [enocom-experiments-304623:us-central1:pg2] Current certificate expiration = 2024-04-15T17:26:49Z
2024/04/15 10:26:49 [enocom-experiments-304623:us-central1:pg2] Connection info refresh operation scheduled at 2024-04-15T17:22:49Z (now + 56m0s)
2024/04/15 10:26:49 [enocom-experiments-304623:us-central1:pg2] Listening on 127.0.0.1:5432
2024/04/15 10:26:49 The proxy has started successfully and is ready for new connections!


2024/04/15 10:26:58 [enocom-experiments-304623:us-central1:pg2] Accepted connection from 127.0.0.1:47630
2024/04/15 10:26:58 [enocom-experiments-304623:us-central1:pg2] Connection info refresh operation started
2024/04/15 10:26:58 [enocom-experiments-304623:us-central1:pg2] Connection info refresh operation complete
2024/04/15 10:26:58 [enocom-experiments-304623:us-central1:pg2] Current certificate expiration = 2024-04-15T17:26:57Z
2024/04/15 10:26:58 [enocom-experiments-304623:us-central1:pg2] Connection info refresh operation scheduled at 2024-04-15T17:22:57Z (now + 56m0s)
2024/04/15 10:26:58 [enocom-experiments-304623:us-central1:pg2] Now = 2024-04-15T16:26:58Z, Current cert expiration = 2024-04-15T17:26:57Z
2024/04/15 10:26:58 [enocom-experiments-304623:us-central1:pg2] Cert is valid = true
2024/04/15 10:26:58 [enocom-experiments-304623:us-central1:pg2] Dialing 35.238.106.212:3307
```

This is happening because the Proxy [calls EngineVersion](https://github.com/GoogleCloudPlatform/cloud-sql-proxy/blob/main/internal/proxy/proxy.go#L520-L521) to warm the cache, but EngineVersion doesn't support a way to specify [if IAM Authn is used or not](https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/blob/a66d85bc8aafb91bcd60c23cba1cff591e3b4003/dialer.go#L383), and so assumes it's not in use (the nil argument).

When the call to connection info occurs, it enables IAM auth, and causes the existing refresh to be canceled.

enocom:
When callers (like the Cloud SQL Proxy) warmup the background refresh
with EngineVersion, the initial refresh operation starts with IAM Authn
disabled. Then when a caller connects with IAM authentication, the
existing refresh is invalidated (because it doesn't include the client's
OAuth2 token). In effect, two refresh operations are completed when the
dialer could have just run one initially.

This commit ensures calls to EngineVersion respect the dialer's global
IAM authentication setting. If IAM authentication is enabled at the
dialer level, then EngineVersion will ensure the refresh operation uses
IAM authentication. And only one refresh operation occurs between warmup
and first connection.

In cases where a dialer is initialized without IAM authentication, but
then a call to dial requests IAM authentication, a second refresh is
unavoidable. This seems to be an uncommon enough use case that it is an
acceptable tradeoff given how IAM authentication is tightly coupled with
the client certificate refresh.

Separately, when Cloud SQL Proxy invocations start the Proxy with the
--token flag in combination with IAM authentication, the underlying
token does not have a corresponding refresh token and so cannot be
refreshed. As a result, when the double calls occur (as described
above), there is a third refresh attempt started because the token has a
missing expiration field (there is only AccessToken, no RefreshToken, or
Expiry).

The dialer sees the missing expiration as an expired client certificate
and immediately starts a new refresh. But because the dialer has already
consumed two attempts, the rate limiter (2 initial attempts, then
30s/attempt) forces the client to wait 30s before connecting.

This commit ensures that situation will not happen by using the correct
client certificate expiration and not the invalid token expiration. For
cases where the Cloud SQL Proxy configures the dialer with the --token
flag (and no refresh token), the dialer will always default to using the
client certificate's expiration. This means the refresh cycle will fail
once the token expires.

Fixes #771

### DIFF
--- a/internal/cloudsql/instance.go
+++ b/internal/cloudsql/instance.go
@@ -393,8 +393,13 @@ func (i *RefreshAheadCache) scheduleRefresh(d time.Duration) *refreshOperation {
 				nil,
 			)
 		} else {
+			var useIAMAuthN bool
+			i.mu.Lock()
+			useIAMAuthN = i.useIAMAuthNDial
+			i.mu.Unlock()
 			r.result, r.err = i.r.ConnectionInfo(
-				ctx, i.connName, i.key, i.useIAMAuthNDial)
+				ctx, i.connName, i.key, useIAMAuthN,
+			)
 		}
 		switch r.err {
 		case nil:
--- a/dialer.go
+++ b/dialer.go
@@ -389,7 +389,7 @@ func (d *Dialer) EngineVersion(ctx context.Context, icn string) (string, error) 
 	if err != nil {
 		return "", err
 	}
-	i := d.connectionInfoCache(ctx, cn, nil)
+	i := d.connectionInfoCache(ctx, cn, &d.defaultDialConfig.useIAMAuthN)
 	ci, err := i.ConnectionInfo(ctx)
 	if err != nil {
 		return "", err
--- a/dialer_test.go
+++ b/dialer_test.go
@@ -396,6 +396,38 @@ func TestDialerEngineVersion(t *testing.T) {
 	}
 }
 
+// When Auto IAM AuthN is enabled, EngineVersion should warm the cache with a
+// client certificate with Auto IAM AuthN enabled.
+func TestEngineVersionAvoidsDuplicateRefreshWithIAMAuthN(t *testing.T) {
+	inst := mock.NewFakeCSQLInstance(
+		"my-project", "my-region", "my-instance",
+	)
+	d := setupDialer(t, setupConfig{
+		testInstance: inst,
+		dialerOptions: []Option{
+			WithIAMAuthN(), WithIAMAuthNTokenSources(
+				mock.EmptyTokenSource{},
+				mock.EmptyTokenSource{},
+			),
+		},
+		reqs: []*mock.Request{
+			// There should only be two API requests
+			mock.InstanceGetSuccess(inst, 1),
+			mock.CreateEphemeralSuccess(inst, 1),
+		},
+	})
+
+	_, err := d.EngineVersion(context.Background(), inst.String())
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	testSuccessfulDial(
+		context.Background(), t, d,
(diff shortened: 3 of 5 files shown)


internal/cloudsql/instance.go
```diff
@@ -393,8 +393,13 @@ func (i *RefreshAheadCache) scheduleRefresh(d time.Duration) *refreshOperation {
 				nil,
 			)
 		} else {
+			var useIAMAuthN bool
+			i.mu.Lock()
+			useIAMAuthN = i.useIAMAuthNDial
+			i.mu.Unlock()
```
jackwotherspoon, line 399: why do we need a lock around this?

jackwotherspoon (approved):
LGTM, great PR description 🤩 👏

internal/cloudsql/instance.go
enocom: During my testing, this triggered a race condition between the read here and the write in UpdateRefresh. I wrapped it with a lock to prevent any data races in the case that a refresh is scheduled at the same time a refresh was being updated.

PR #810, opened 2024-05-22
chore: make short tests faster

enocom:
The test cases covering failure behavior were all inadvertently using exponential backoff in test runs, causing unnecessary slowdowns. This commit disables exponential backoff in tests by having the mock API server return 400 status codes instead of 501s when a request is not recognized. As a result, the exponential backoff logic does not trigger.

This change reduces the time taken by ~75% to about ~5s.


jackwotherspoon (approved):
❤️ 🤩

PR #812, opened 2024-05-28
feat: invalidate cache on failed IP lookup

### LINKED ISSUE #780: Invalidate cache on failed IP lookup
In https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/pull/768, we removed an implicit behavior of the Go Connector. If a dial attempt requests a non-existent IP type (e.g., client asks for public IP on a private IP only instance), the Connector would invalidate the cache. But with the cleanup PR, we removed that implicit behavior.

In some cases, it might be useful to have this behavior. For example, if a caller starts the Go Connector and tries to connect to a public IP and then later configures public IP, there is no need for a restart.

We made that change in the AlloyDB Go Connector mostly because some internal tests depend on that behavior. See https://github.com/GoogleCloudPlatform/alloydb-go-connector/pull/555.

This is for porting the same behavior here.

enocom:
In #768, we removed an implicit behavior of the Go Connector. If a dial attempt requests a non-existent IP type (e.g., client asks for public IP on a private IP only instance), the Connector would invalidate the cache. But with the cleanup PR, we removed that implicit behavior.

In some cases, it might be useful to have this behavior. For example, if a caller starts the Go Connector and tries to connect to a public IP and then later configures public IP, there is no need for a restart.

We made that change in the AlloyDB Go Connector mostly because some internal tests depend on that behavior. See
GoogleCloudPlatform/alloydb-go-connector#555.

Fixes #780

### DIFF
--- a/dialer.go
+++ b/dialer.go
@@ -275,12 +275,7 @@ func (d *Dialer) Dial(ctx context.Context, icn string, opts ...DialOption) (conn
 	c := d.connectionInfoCache(ctx, cn, &cfg.useIAMAuthN)
 	ci, err := c.ConnectionInfo(ctx)
 	if err != nil {
-		d.lock.Lock()
-		defer d.lock.Unlock()
-		d.logger.Debugf(ctx, "[%v] Removing connection info from cache", cn.String())
-		// Stop all background refreshes
-		c.Close()
-		delete(d.cache, cn)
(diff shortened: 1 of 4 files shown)


jackwotherspoon (approved):
LGTM ✅

PR #824, opened 2024-06-05
fix: ensure connection count is correctly reported

enocom:
A recent refactor broke the open connections metric and caused all counts to be reported as either 1 or 0. This commit updates the test to protect against making the same mistake and fixes the original problem (using value semanatics and not pointer semantics).


jackwotherspoon (approved):
LGTM

PR #827, opened 2024-06-06
fix: invalidate cache on failed `Warmup` and `EngineVersion`

### LINKED ISSUE #778: Warmup and EngineVersion don't close cache on failed lookup
EngineVersion doesn't close the cache on error:

https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/blob/c4fb69f9b7ec32be3bd07fbf399b5dc1ca0da88c/dialer.go#L387-L396

Warmup doesn't close the cache on error:

https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/blob/c4fb69f9b7ec32be3bd07fbf399b5dc1ca0da88c/dialer.go#L403-L412

As a result, if a client connects with a bad instance connection name, missing IAM permissions, no network connectivity etc, the background refresh will continue running regardless.

Instead, we should close the cache and delete it from the shared map.

jackwotherspoon:
Invalidate the cache and stop background refreshes on failed `Warmup` and `EngineVersion` calls.

Fixes #778

### DIFF
--- a/dialer.go
+++ b/dialer.go
@@ -401,9 +401,10 @@ func (d *Dialer) EngineVersion(ctx context.Context, icn string) (string, error) 
 	if err != nil {
 		return "", err
 	}
-	i := d.connectionInfoCache(ctx, cn, &d.defaultDialConfig.useIAMAuthN)
-	ci, err := i.ConnectionInfo(ctx)
+	c := d.connectionInfoCache(ctx, cn, nil)
+	ci, err := c.ConnectionInfo(ctx)
 	if err != nil {
+		d.removeCached(ctx, cn, c, err)
 		return "", err
 	}
 	return ci.DBVersion, nil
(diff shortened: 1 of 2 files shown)


dialer.go
```diff
@@ -456,7 +457,10 @@ func (d *Dialer) Warmup(ctx context.Context, icn string, opts ...DialOption) err
 	for _, opt := range opts {
 		opt(&cfg)
 	}
-	_, err = d.connectionInfoCache(ctx, cn, &cfg.useIAMAuthN)
+	c, err := d.connectionInfoCache(ctx, cn, &cfg.useIAMAuthN)
```
enocom, line 460: The error return from connection info cache is really only interesting if the key generator fails. If an IAM permission is missing, the call to `d.connectionInfoCache` won't return an error until someone asks for `ConnectionInfo`. So as written there's no way to properly close the internal cache.

What if we just replaced these internals with a call to `EngineVersion`? That would solve all the problems above and make use of the error checking you've added.

enocom: I see the API signatures don't support this, so we'll have to just match the internal implementation between EngineVersion and Warmup.

jackwotherspoon: As discussed offline, I have made `Warmup` call `ConnectionInfo` as there is no other way to invalidate without breaking the API.

enocom (approved):

PR #843, opened 2024-07-16
feat: Support configure connections using DNS

hessjcg:
The dialer may be configured to use a DNS name to look up the instance 
name instead of configuring the connector with the instance name directly. 

Add a DNS TXT record for the Cloud SQL instance to a **private** DNS server 
or a private Google Cloud DNS Zone used by your application.  For example:

- Record type: `TXT` 
- Name: `prod-db.mycompany.example.com` – This is the domain name used by the application
- Value: `my-project:region:my-instance` – This is the instance connection name

Configure the dialer with the `cloudsqlconn.WithDNSResolver()` option. 

Open a database connection using the DNS name: 

```go
db, err := sql.Open(
    "cloudsql-mysql",
    "myuser:mypass@cloudsql-mysql(prod-db.mycompany.example.com)/mydb",
)
```

Part of #842

### DIFF
--- a/README.md
+++ b/README.md
@@ -265,28 +265,30 @@ the DNS name.
 Adapting the MySQL + database/sql example above:
 
 ```go
-import (
-    "database/sql"
+package main
 
-    "cloud.google.com/go/cloudsqlconn"
-    "cloud.google.com/go/cloudsqlconn/mysql/mysql"
+import (
+	"database/sql"
+
+	"cloud.google.com/go/cloudsqlconn"
+	"cloud.google.com/go/cloudsqlconn/mysql/mysql"
 )
 
 func connect() {
-    cleanup, err := mysql.RegisterDriver("cloudsql-mysql",
-			cloudsqlconn.WithDnsResolver(),
-			cloudsqlconn.WithCredentialsFile("key.json"))
-    if err != nil {
-        // ... handle error
-    }
-    // call cleanup when you're done with the database connection
-    defer cleanup()
+	cleanup, err := mysql.RegisterDriver("cloudsql-mysql",
+		cloudsqlconn.WithDNSResolver(),
+		cloudsqlconn.WithCredentialsFile("key.json"))
+	if err != nil {
+		// ... handle error
+	}
+	// call cleanup when you're done with the database connection
+	defer cleanup()
 
-    db, err := sql.Open(
-        "cloudsql-mysql",
-        "myuser:mypass@cloudsql-mysql(prod-db.mycompany.example.com)/mydb",
-    )
-    // ... etc
+	db, err := sql.Open(
+		"cloudsql-mysql",
+		"myuser:mypass@cloudsql-mysql(prod-db.mycompany.example.com)/mydb",
+	)
+	// ... etc
 }
 ```
 
--- a/README.md
+++ b/README.md
@@ -234,6 +234,63 @@ func connect() {
     // ... etc
 }
 ```
+### Using DNS to identify an instance
+
+The connector can be configured to use DNS to look up an instance. This would
+allow you to configure your application to connect to a database instance, and
+centrally configure which instance in your DNS zone.
+
+#### Configure your DNS Records
+
+Add a DNS SRV record for the Cloud SQL instance to a **private** DNS server 
+or a private Google Cloud DNS Zone used by your application. 
+
+**Note:** You are strongly discouraged from adding DNS records for your 
+Cloud SQL instances to a public DNS server. This would allow anyone on the
+internet to discover the Cloud SQL instance name. 
+
+For example: suppose you wanted to use the domain name 
+`prod-db.mycompany.example.com` to connect to your database instance 
+`my-project:region:my-instance`. 
+
+- Record type: `SRV` 
+- Name: `prod-db.mycompany.example.com` – This is the domain name used by the application
+- Target: `my-project:region:my-instance` – This is the instance name
+- Port: `3307` – always use port 3307
+- Priority: `0` – always use priority 0 
+- Weight: `1` - always use weight 1
+
+#### Configure the connector
+
+Configure the connector as described above, replacing the conenctor ID with
+the DNS name. 
+
+Adapting the MySQL + database/sql example above:
+
+```go
+import (
+    "database/sql"
+
+    "cloud.google.com/go/cloudsqlconn"
+    "cloud.google.com/go/cloudsqlconn/mysql/mysql"
+)
+
+func connect() {
+    cleanup, err := mysql.RegisterDriver("cloudsql-mysql", cloudsqlconn.WithCredentialsFile("key.json"))
+    if err != nil {
+        // ... handle error
+    }
+    // call cleanup when you're done with the database connection
+    defer cleanup()
+
+    db, err := sql.Open(
+        "cloudsql-mysql",
+        "myuser:mypass@cloudsql-mysql(prod-db.mycompany.example.com)/mydb",
+    )
+    // ... etc
+}
+```
+
 
 ### Using Options
 
--- a/README.md
+++ b/README.md
@@ -242,7 +242,7 @@ centrally configure which instance in your DNS zone.
 
 #### Configure your DNS Records
 
-Add a DNS SRV record for the Cloud SQL instance to a **private** DNS server 
+Add a DNS TXT record for the Cloud SQL instance to a **private** DNS server 
 or a private Google Cloud DNS Zone used by your application. 
 
 **Note:** You are strongly discouraged from adding DNS records for your 
@@ -251,14 +251,11 @@ internet to discover the Cloud SQL instance name.
 
 For example: suppose you wanted to use the domain name 
 `prod-db.mycompany.example.com` to connect to your database instance 
-`my-project:region:my-instance`. 
+`my-project:region:my-instance`. You would create the following DNS record: 
 
-- Record type: `SRV` 
+- Record type: `TXT` 
 - Name: `prod-db.mycompany.example.com` – This is the domain name used by the application
-- Target: `my-project:region:my-instance` – This is the instance name
-- Port: `3307` – always use port 3307
-- Priority: `0` – always use priority 0 
-- Weight: `1` - always use weight 1
+- Value: `my-project:region:my-instance` – This is the instance name
 
 #### Configure the connector
 
@@ -276,7 +273,9 @@ import (
 )
 
 func connect() {
-    cleanup, err := mysql.RegisterDriver("cloudsql-mysql", cloudsqlconn.WithCredentialsFile("key.json"))
+    cleanup, err := mysql.RegisterDriver("cloudsql-mysql",
+			cloudsqlconn.WithDnsResolver(),
+			cloudsqlconn.WithCredentialsFile("key.json"))
     if err != nil {
         // ... handle error
     }
--- a/dialer.go
+++ b/dialer.go
@@ -155,7 +155,7 @@ type Dialer struct {
 	iamTokenSource oauth2.TokenSource
 
 	// resolver converts instance names into DNS names.
-	resolver InstanceConnectionNameResolver
+	resolver instance.ConnectionNameResolver
 }
 
 var (
@@ -256,7 +256,7 @@ func NewDialer(ctx context.Context, opts ...Option) (*Dialer, error) {
 	if err != nil {
 		return nil, err
 	}
-	var r InstanceConnectionNameResolver = cloudsql.DefaultInstanceConnectionNameResolver
+	var r instance.ConnectionNameResolver = cloudsql.DefaultResolver
 	if cfg.resolver != nil {
 		r = cfg.resolver
 	}
@@ -297,7 +297,7 @@ func (d *Dialer) Dial(ctx context.Context, icn string, opts ...DialOption) (conn
 		go trace.RecordDialError(context.Background(), icn, d.dialerID, err)
 		endDial(err)
 	}()
-	cn, err := d.resolver.Lookup(ctx, icn)
+	cn, err := d.resolver.Resolve(ctx, icn)
 	if err != nil {
 		return nil, err
 	}
@@ -438,7 +438,7 @@ func validClientCert(
 // the instance:
 // https://cloud.google.com/sql/docs/mysql/admin-api/rest/v1beta4/SqlDatabaseVersion
 func (d *Dialer) EngineVersion(ctx context.Context, icn string) (string, error) {
-	cn, err := d.resolver.Lookup(ctx, icn)
+	cn, err := d.resolver.Resolve(ctx, icn)
 	if err != nil {
 		return "", err
 	}
@@ -458,7 +458,7 @@ func (d *Dialer) EngineVersion(ctx context.Context, icn string) (string, error) 
 // Use Warmup to start the refresh process early if you don't know when you'll
 // need to call "Dial".
 func (d *Dialer) Warmup(ctx context.Context, icn string, opts ...DialOption) error {
-	cn, err := d.resolver.Lookup(ctx, icn)
+	cn, err := d.resolver.Resolve(ctx, icn)
 	if err != nil {
 		return err
 	}
--- a/dialer.go
+++ b/dialer.go
@@ -153,6 +153,9 @@ type Dialer struct {
 
 	// iamTokenSource supplies the OAuth2 token used for IAM DB Authn.
 	iamTokenSource oauth2.TokenSource
+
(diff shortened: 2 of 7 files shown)


dialer.go
```diff
@@ -253,6 +258,13 @@ func NewDialer(ctx context.Context, opts ...Option) (*Dialer, error) {
 	if err != nil {
 		return nil, err
 	}
+	var r NetResolver
```
enocom, line 261:
``` golang
r := net.DefaultResolver
if cfg.resolver != nil {
  r = cfg.resolver
}
```

https://go.dev/wiki/CodeReviewComments#indent-error-flow

dialer.go
```diff
@@ -288,9 +315,14 @@ func (d *Dialer) Dial(ctx context.Context, icn string, opts ...DialOption) (conn
 		go trace.RecordDialError(context.Background(), icn, d.dialerID, err)
 		endDial(err)
 	}()
-	cn, err := instance.ParseConnName(icn)
+	cn, err := d.resolveInstanceName(ctx, icn)
 	if err != nil {
-		return nil, err
+		// The connection name was not project:region:instance
```
enocom, line 320: I think we don't need this block of code since it's part of `resolveInstanceName`?

dialer.go
```diff
@@ -565,3 +597,43 @@ func (d *Dialer) connectionInfoCache(
 
 	return c, nil
 }
+
+func (d *Dialer) queryDNS(ctx context.Context, domainName string) (instance.ConnName, error) {
```
enocom, line 601: Since this is a non-trivial method, let's add a godoc (even though it's private) for future readers.

dialer.go
```diff
@@ -565,3 +597,43 @@ func (d *Dialer) connectionInfoCache(
...
+		if parseErr != nil {
+			perr = fmt.Errorf("unable to parse SRV for %s: %v", domainName, parseErr)
+			continue
+		}
+		return cn, nil
+	}
+
+	// If resolve failed and no records were found, return the error.
+	if err != nil {
+		return instance.ConnName{}, fmt.Errorf("unable to resolve SRV record for %s: %v", domainName, err)
```
enocom, line 629: %q might be nice here for the SRV record -- it will wrap the value in quotes.

options.go
```diff
@@ -34,6 +34,13 @@ import (
 // An Option is an option for configuring a Dialer.
 type Option func(d *dialerConfig)
 
+// NetResolver groups the methods on net.Resolver that are used by the DNS
```
enocom, line 37: If we're adding something to the public API for tests only, we should change our approach (maybe by patching an unexported field).

If on the other hand we think this is a useful interface for callers, we should remove "This allows the default net.Resolver instance to be overridden from tests" and elaborate on why a person might want to provide their own interface.

enocom: Also, nit if we're keeping this interface let's move it down to just above where it's used.

README.md
```diff
@@ -234,6 +234,59 @@ func connect() {
...
 ```
+### Using DNS to identify an instance
+
+The connector can be configured to use DNS to look up an instance. This would
+allow you to configure your application to connect to a database instance, and
+centrally configure which instance in your DNS zone.
+
+#### Configure your DNS Records
+
+Add a DNS SRV record for the Cloud SQL instance to the DNS used by your
```
enocom, line 245: People will do this for public IP. We should warn them that it leaks PII if they do that. This should be recommended for private contexts only.

enocom (changes requested):

dialer.go
hessjcg: Method comment added.

dialer.go

options.go
hessjcg: Moved to just above the `WithResolver()` function. 

I think this makes sense as a public API. I can imagine a customer keeping their DNS entries in a separate, private DNS server. Then they would need to configure the connector to use that DNS server.

options.go
```diff
@@ -234,6 +235,24 @@ func WithIAMAuthN() Option {
 	}
 }
 
+// NetResolver groups the methods on net.Resolver that are used by the DNS
+// resolver implementation. This allows an application to replace the default
+// net.DefaultResolver with a custom implementation. For example: the
+// application may need to connect to a specific DNS server using a specially
+// configured instance of net.Resolver.
+type NetResolver interface {
+	LookupSRV(ctx context.Context, service, proto, name string) (cname string, addrs []*net.SRV, err error)
```
enocom, line 244: The more I think about this, the more I think this is the wrong interface. What we want generically is a way to go from some name (DNS record or key value store path or whatever the user wants) to an instance connection name. This would be more future proof for when we want to support alternate resolution techniques

``` golang
type InstanceConnectionNameResolver interface {
    Resolve(ctx context.Context, name string) (string, error) 
}
```

Then we could provide a default DNS resolve that does the equivalent of the SRV lookup. I think this is a small change, but will unlock many more use cases (especially as we discover the shortcomings of DNS polling).

enocom (changes requested):

options.go
hessjcg: Good insight, I agree. The Java implementation started heading in this direction too. I will update the PR accordingly.

hessjcg (commented):
I created a new public interface `InstanceConnectionNameResolver` and an internal implementation called `DNSInstanceConnectionNameResolver`.

README.md
```diff
@@ -234,6 +234,63 @@ func connect() {
...
+#### Configure your DNS Records
+
+Add a DNS SRV record for the Cloud SQL instance to a **private** DNS server 
+or a private Google Cloud DNS Zone used by your application. 
+
+**Note:** You are strongly discouraged from adding DNS records for your 
+Cloud SQL instances to a public DNS server. This would allow anyone on the
+internet to discover the Cloud SQL instance name. 
+
+For example: suppose you wanted to use the domain name
```
enocom, line 252: "For example, suppose ... to your database instance `my-project:region:my-instance`, you would create the following DNS record:

README.md
```diff
@@ -234,6 +234,63 @@ func connect() {
...
+```go
+import (
+    "database/sql"
+
+    "cloud.google.com/go/cloudsqlconn"
+    "cloud.google.com/go/cloudsqlconn/mysql/mysql"
+)
+
+func connect() {
+    cleanup, err := mysql.RegisterDriver("cloudsql-mysql", cloudsqlconn.WithCredentialsFile("key.json"))
```
enocom, line 279: I think we should have an explicit opt-in to this DNS feature, e.g.,

`WithDNSResolution()`

wdyt?

dialer.go
```diff
@@ -253,6 +256,11 @@ func NewDialer(ctx context.Context, opts ...Option) (*Dialer, error) {
 	if err != nil {
 		return nil, err
 	}
+	var r InstanceConnectionNameResolver = cloudsql.DefaultInstanceConnectionNameResolver
```
enocom, line 259: What if we used a shorter name, e.g., `DefaultResolver`? We have a public `instance` package where we could put this, the interface, and the SRVResolver.

options.go
```diff
@@ -234,6 +236,21 @@ func WithIAMAuthN() Option {
 	}
 }
 
+// InstanceConnectionNameResolver resolves  This allows an application to replace the default
+// DNSInstanceConnectionNameResolver with a custom implementation.
+type InstanceConnectionNameResolver interface {
```
enocom, line 241: This should probably go in the `instance` package next to the code that parses instance connection names. And that way we could just call it `Resolver`.

options.go
```diff
@@ -234,6 +236,21 @@ func WithIAMAuthN() Option {
...
+	Lookup(ctx context.Context, name string) (instanceName instance.ConnName, err error)
```
enocom, line 242: Resolver -> Resolve would be more Go-like.

internal/cloudsql/resolver.go
```diff
@@ -0,0 +1,118 @@
...
+var DefaultInstanceConnectionNameResolver = &DNSInstanceConnectionNameResolver{
+	dnsResolver: net.DefaultResolver,
+}
+
+// netResolver groups the methods on net.Resolver that are used by the DNS
+// resolver implementation. This allows an application to replace the default
+// net.DefaultResolver with a custom implementation. For example: the
+// application may need to connect to a specific DNS server using a specially
+// configured instance of net.Resolver.
+type netResolver interface {
```
enocom, line 38: Could we remove the duplicate interface and test this e2e? Or alternatively, what would it take to assemble a mock DNS server in unit tests?

enocom (changes requested):

README.md
hessjcg: Updated. Switching to TXT records.

internal/cloudsql/resolver.go
hessjcg: I think we need a mock for practicality. We need to write test cases covering our code when the DNS records are misconfigured. Also, we need to do this across all our supported connector languages. Mocking the DNS response is the most reliably way to implement these test cases.

dialer.go
hessjcg: Yes. See review comment.

options.go
hessjcg: Yes. See review comment.

options.go
hessjcg: Yes. See review comment.

hessjcg (commented):
I did a bit of refactoring as recommended. 

I moved the `ConnectionNameResolver` interface into the exported `instance` package and renamed it's method from `Lookup()` to `Resolve()`

I created two option methods: `WithResolver(r ConnectionNameResolver)` and `WithDnsResolver()`. 

I configured the default behavior of the dialer to only accept instance names.

README.md
```diff
@@ -234,6 +234,62 @@ func connect() {
...
+import (
+    "database/sql"
+
+    "cloud.google.com/go/cloudsqlconn"
+    "cloud.google.com/go/cloudsqlconn/mysql/mysql"
+)
+
+func connect() {
+    cleanup, err := mysql.RegisterDriver("cloudsql-mysql",
+			cloudsqlconn.WithDnsResolver(),
```
enocom, line 277: WithDNSResolver()

Also, let's do the equivalent of gofmt on this sample code.

dialer_test.go
```diff
@@ -1016,3 +1016,64 @@ func TestDialerInitializesLazyCache(t *testing.T) {
 		t.Fatalf("dialer was initialized with non-lazy type: %T", tt)
 	}
 }
+
+type fakeResolver struct{}
+
+func (r *fakeResolver) Resolve(_ context.Context, name string) (instance.ConnName, error) {
+	// For TestDialerSuccessfullyDialsDnsTxtRecord
+	if name == "db.example.com" {
```
enocom, line 1024: Instead of relying on hardcoded values for particular tests, you could set fields in `fakeResolver` which makes this more future proof and easy to extend:

``` golang

type fakeResolve struct {
  instances map[string]struct{
    connName instance.ConnName
    err error
  }
}
```

And then this implementation becomes:

``` golang
func (r *fakeResolver) Resolve(_ context.Context, name string) (instance.ConnName, error) {
  icn, ok := r.instances[name]
  if !ok {
    return instance.ConnName{}, errors.New("not found")
  }
  return icn, nil
}
```

dialer_test.go
```diff
@@ -1016,3 +1016,64 @@ func TestDialerInitializesLazyCache(t *testing.T) {
...
+			mock.InstanceGetSuccess(inst, 1),
+			mock.CreateEphemeralSuccess(inst, 1),
+		},
+		dialerOptions: []Option{
+			WithTokenSource(mock.EmptyTokenSource{}),
+			WithResolver(&fakeResolver{}),
+		},
+	})
+
+	// Target has a trailing '.'
```
enocom, line 1050: stale?

instance/conn_name.go
```diff
@@ -74,3 +75,9 @@ func ParseConnName(cn string) (ConnName, error) {
 	}
 	return c, nil
 }
+
+// ConnectionNameResolver resolves  This allows an application to replace the default
```
enocom, line 79: Let's clean up this to make is a complete sentence starting with the interface name.

instance/conn_name.go
```diff
@@ -74,3 +75,9 @@ func ParseConnName(cn string) (ConnName, error) {
...
+// DNSInstanceConnectionNameResolver with a custom implementation.
```
enocom, line 80: Maybe just "replace the default instance connection name resolution..."

instance/conn_name.go
```diff
@@ -74,3 +75,9 @@ func ParseConnName(cn string) (ConnName, error) {
...
+type ConnectionNameResolver interface {
+	Resolve(ctx context.Context, name string) (instanceName ConnName, err error)
```
enocom, line 82: The named return values are unnecessary. All consider adding a godoc to `Resolve` to explain what this should do?

internal/cloudsql/resolver_test.go
```diff
@@ -0,0 +1,77 @@
...
+	"testing"
+
+	"cloud.google.com/go/cloudsqlconn/instance"
+)
+
+type fakeResolver struct{}
+
+func (r *fakeResolver) LookupTXT(_ context.Context, name string) (addrs []string, err error) {
+	// For TestDialerSuccessfullyDialsDnsTXTRecord
+	if name == "db.example.com" {
```
enocom, line 30: Same thing here.

options.go
```diff
@@ -234,6 +236,24 @@ func WithIAMAuthN() Option {
...
+// instance.
+func WithResolver(r instance.ConnectionNameResolver) Option {
+	return func(d *dialerConfig) {
+		d.resolver = r
+	}
+}
+
+// WithDNSResolver replaces the default resolver (which only resolves instance
+// names) with the DNSResolver, which will attempt to first parse the instance
+// name, and then will attempt to resolve the DNS TXT record.
```
enocom, line 250: Could we elaborate here on how this DNS TXT record works like we have in the README? Many Go developers won't see the README and will go straight to to Go doc's.

internal/cloudsql/resolver.go
```diff
@@ -0,0 +1,123 @@
...
+	"fmt"
+	"net"
+	"sort"
+
+	"cloud.google.com/go/cloudsqlconn/instance"
+)
+
+// DNSResolver uses the default net.Resolver to find
+// TXT records containing an instance name for a DNS record.
+var DNSResolver = &DNSInstanceConnectionNameResolver{
```
enocom, line 28: Does this need to be exported?

enocom (changes requested):
Looking good. A few more changes to make here.

README.md
hessjcg: Fixed

dialer_test.go
hessjcg: This seems excessive. After simplifying obsolete test cases, `fakeResolver` only ever needs to resolve one domain name.

dialer_test.go
hessjcg: Stale. removed.

instance/conn_name.go
hessjcg: Fixed.

instance/conn_name.go
hessjcg: fixed.

instance/conn_name.go
hessjcg: fixed.

internal/cloudsql/resolver.go
hessjcg: Yes, it's used in /options.go

internal/cloudsql/resolver_test.go
hessjcg: Fixed.

options.go
hessjcg: Description added.

dialer_test.go
enocom: I think it helps test readability. Otherwise, a person has to go searching for what's special about `db.example.com`.

If we're not supporting multiple instances anymore this also works:

``` golang
WithResolver(&fakeResolver{validName: "db.example.com"}),
```

``` golang
type fakeResolve struct {
  validName string
}

func (r *fakeResolver) Resolve(_ context.Context, name string) (instance.ConnName, error) {
  if name != r.validName {
    // return error
  }
  // return connection name
}
```

The point is to make the fake extensible -- not just for this use, but for future use.

enocom: Actually this is what you do below in the resolver test -- let's do the same thing here.

hessjcg: Done.

enocom (approved):
This will be a well-liked feature.

PR #844, opened 2024-07-17
feat: Automatically reset DNS-configured connections on DNS change

hessjcg:
When a connection is configured using a DNS domain name, the connector will poll the DNS TXT record every 30 seconds. If the value of the DNS record changes, the connector will close all connections to the old instance, and direct new connections to the updated instance. 

Part of #842

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,7 +234,8 @@ func connect() {
     // ... etc
 }
 ```
-### Using DNS to identify an instance
+
+### Using DNS domain names to identify instances
 
 The connector can be configured to use DNS to look up an instance. This would
 allow you to configure your application to connect to a database instance, and
@@ -291,6 +292,41 @@ func connect() {
 	// ... etc
 }
 ```
+
+### Automatic fail-over using DNS domain names
+
+When the connector is configured using a domain name, the connector will 
+periodically check if the DNS record for an instance changes. When the connector 
+detects that the domain name refers to a different instance, the connector will
+close all open connections to the old instance. Subsequent connection attempts
+will be directed to the new instance. 
+
+For example: suppose application is configured to connect using the
+domain name `prod-db.mycompany.example.com`. Initially the corporate DNS 
+zone has a TXT record with the value `my-project:region:my-instance`. The
+application establishes connections to the `my-project:region:my-instance` 
+Cloud SQL instance. 
+
+Then, to reconfigure the application using a different database
+instance: `my-project:other-region:my-instance-2`. You update the DNS record
+for `prod-db.mycompany.example.com` with the target 
+`my-project:other-region:my-instance-2`
+
+The connector inside the application detects the change to this
+DNS entry. Now, when the application connects to its database using the 
+domain name `prod-db.mycompany.example.com`, it will connect to the
+`my-project:other-region:my-instance-2` Cloud SQL instance. 
+
+The connector will automatically close all existing connections to
+`my-project:region:my-instance`. This will force the connection pools to 
+establish new connections. Also, it may cause database queries in progress 
+to fail. 
+
+The connector will poll for changes to the DNS name every 30 seconds by default.
+You may configure the frequency of the connections using the option 
+`WithFailoverPeriod(d time.Duration)`. When this is set to 0, the connector will
+disable polling and only check if the DNS record changed when it is
+creating a new connection. 
 
 
 ### Using Options
--- a/dialer.go
+++ b/dialer.go
@@ -107,39 +107,8 @@ type connectionInfoCache interface {
 	ConnectionInfo(context.Context) (cloudsql.ConnectionInfo, error)
 	UpdateRefresh(*bool)
 	ForceRefresh()
+	UseIAMAuthN() bool
 	io.Closer
-}
-
-// monitoredCache is a wrapper around a connectionInfoCache that tracks the
-// number of connections to the associated instance.
-type monitoredCache struct {
-	openConnsCount *uint64
-
-	connectionInfoCache
-}
-
-func (c *monitoredCache) Close() error {
-	return c.connectionInfoCache.Close()
-}
-
-func (c *monitoredCache) ForceRefresh() {
-	if c == nil || c.connectionInfoCache == nil {
-		return
-	}
-	c.connectionInfoCache.ForceRefresh()
-}
-
-func (c *monitoredCache) UpdateRefresh(b *bool) {
-	if c == nil || c.connectionInfoCache == nil {
-		return
-	}
-	c.connectionInfoCache.UpdateRefresh(b)
-}
-func (c *monitoredCache) ConnectionInfo(ctx context.Context) (cloudsql.ConnectionInfo, error) {
-	if c == nil || c.connectionInfoCache == nil {
-		return cloudsql.ConnectionInfo{}, nil
-	}
-	return c.connectionInfoCache.ConnectionInfo(ctx)
 }
 
 // A Dialer is used to create connections to Cloud SQL instances.
@@ -178,7 +147,8 @@ type Dialer struct {
 	iamTokenSource oauth2.TokenSource
 
 	// resolver converts instance names into DNS names.
-	resolver instance.ConnectionNameResolver
+	resolver       instance.ConnectionNameResolver
+	failoverPeriod time.Duration
 }
 
 var (
@@ -202,6 +172,7 @@ func NewDialer(ctx context.Context, opts ...Option) (*Dialer, error) {
 		logger:          nullLogger{},
 		useragents:      []string{userAgent},
 		serviceUniverse: "googleapis.com",
+		failoverPeriod:  cloudsql.FailoverPeriod,
 	}
 	for _, opt := range opts {
 		opt(cfg)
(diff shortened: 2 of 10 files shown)


dialer.go
```diff
@@ -516,12 +552,94 @@ func (d *Dialer) Close() error {
...
 }
 
+func (d *Dialer) pollDomainNames(ctx context.Context) {
+	type cacheEntry struct {
+		cn    instance.ConnName
+		cache monitoredCache
+	}
+
+	// List all the cache entries created with a domain name
+	d.lock.RLock()
```
enocom, line 571: What does this lock protect? Are we sure we want to have it protect cache entries as well?

dialer.go
```diff
@@ -516,12 +552,94 @@ func (d *Dialer) Close() error {
...
+			continue
+		}
+
+		cacheEntries = append(cacheEntries, cacheEntry{cn: cn, cache: cache})
+	}
+	d.lock.RUnlock()
+
+	for _, entry := range cacheEntries {
+		// Remove closed sockets from cache.openSockets
+		entry.cache.lock.Lock()
```
enocom, line 586: `mu` is a more typical name in Go and avoids the stutter: lock.Lock()

README.md
```diff
@@ -292,6 +293,40 @@ func connect() {
 }
 ```
 
+### Automatic fail-over using DNS domain names
+
+When you configure the connector using a domain name, the connector will 
+periodically check if the DNS record for an instance changes. When connector
```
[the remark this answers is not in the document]
hessjcg: Fixed.

README.md
```diff
@@ -292,6 +293,40 @@ func connect() {
...
+`my-project:other-region:my-instance-2`
+
+The connector inside your application detects the change to this
+DNS entry. Now, when your application connects to its database using the 
+domain name `prod-db.mycompany.example.com`, it will connect to the
+`my-project:other-region:my-instance-2` Cloud SQL instance. 
+
+The connector will automatically close all existing connections to
+`my-project:region:my-instance`. This may cause errors in your application as
+database queries in progress will fail.
```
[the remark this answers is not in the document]
hessjcg: Reworded.

dialer.go
```diff
@@ -275,8 +285,23 @@ func NewDialer(ctx context.Context, opts ...Option) (*Dialer, error) {
...
+	if cfg.failoverPeriod > 0 {
+		d.initFailoverRoutine(ctx, cfg.failoverPeriod)
+	}
+
 	return d, nil
 }
+func (d *Dialer) initFailoverRoutine(ctx context.Context, p time.Duration) {
+	d.domainNameTicker = time.NewTicker(p)
+	go func() {
+		for range d.domainNameTicker.C {
```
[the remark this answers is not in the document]

dialer.go
```diff
@@ -516,12 +552,94 @@ func (d *Dialer) Close() error {
...
+	if d.domainNameTicker != nil {
+		d.domainNameTicker.Stop()
+	}
 	for _, i := range d.cache {
 		i.Close()
 	}
 	return nil
 }
 
+func (d *Dialer) pollDomainNames(ctx context.Context) {
```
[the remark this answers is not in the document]
hessjcg: Yeah, this is a little odd. Dialer's primary thing is to establish connections. But as part of establishing a connection for DNS, it also manages the cache used to lookup ConnectionInfoCache instances. Checking for DNS changes falls under "managing the cache". 

In the Java connector, we separated these responsibilities: `InternalConnectorRegistry` does "manage the cache" while `Connector` does "establish connections." 

Maybe we should split `Dialer` in a similar way.

dialer.go
hessjcg: `d.lock` (which was here before this PR) the contents of `d.cache`. Before this PR, the Dialer only added or removed single keys. Iterating through the whole cache is new behavior required by periodic DNS reconciliation.

`d.lock` only needs to protect the state of the `d.cache`. There is a separate lock on each cache entry `entry.cache.lock` which protects the internal state of each `connectionInfoCache.`

dialer.go
hessjcg: I renamed the mutex fields to `mu`.

dialer.go
```diff
@@ -516,12 +552,94 @@ func (d *Dialer) Close() error {
...
+		if newCn != entry.cn {
+			d.logger.Debugf(ctx, "domain name %s changed from old instance %s to new instance %s",
+				entry.cn.DomainName(), entry.cn.String(), newCn.String())
+
+			d.closeDomainNameChanged(ctx, entry.cn, entry.cache,
+				fmt.Errorf("domain name %s changed from old instance %s to new instance %s",
+					entry.cn.DomainName(), entry.cn.String(), newCn.String()))
+			// preload the new cache entry
+			b := entry.cache.UseIAMAuthN()
+			d.connectionInfoCache(ctx, newCn, &b)
```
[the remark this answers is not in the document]
hessjcg: Yeah, maybe we need pull the "create new cache entry" logic out of `d.connectionInfoCache()` into its own method.

hessjcg (commented):
I updated the code with some refactoring based on Eno's comments. Please take another look.

dialer.go
```diff
@@ -538,62 +580,168 @@ func (d *Dialer) Close() error {
...
+		}
+
+		cacheEntries = append(cacheEntries, cacheEntry{cn: cn, cache: cache})
+	}
+	d.mu.RUnlock()
+
+	for _, entry := range cacheEntries {
+		// Remove closed sockets from cache.openSockets
+		entry.cache.mu.Lock()
+		var newOpenSockets []*instrumentedConn
```
enocom, line 617: nit: this could just be open. The type communicates what open is.

dialer.go
```diff
@@ -520,6 +561,7 @@ func (i *instrumentedConn) Write(b []byte) (int, error) {
 // Close delegates to the underlying net.Conn interface and reports the close
 // to the provided closeFunc only when Close returns no error.
 func (i *instrumentedConn) Close() error {
+	i.closed = true
```
enocom, line 564: This isn't thread safe since we have a read happening below on this value.

dialer.go
```diff
@@ -538,62 +580,168 @@ func (d *Dialer) Close() error {
...
+		}
+		entry.cache.openSockets = newOpenSockets
+		entry.cache.mu.Unlock()
+
+		// Resolve the domain name.
+		newCn, err := d.resolver.Resolve(ctx, entry.cn.DomainName())
+
+		// the domain name no longer resolves to a valid instance
+		if err != nil {
+			d.logger.Debugf(ctx, "[failover] unable to resolve DNS for instance %s: %v", entry.cn.DomainName(), err)
```
enocom, line 631: We should have a `continue` after this log statement.

dialer.go
```diff
@@ -115,15 +116,19 @@ type connectionInfoCache interface {
 type monitoredCache struct {
 	openConns *uint64
 
+	mu          sync.Mutex
+	openSockets []*instrumentedConn
```
enocom, line 120: The underlying Go interface is net.Conn. Let's use the same abstraction and call this `openConns`.

dialer.go
```diff
@@ -538,62 +580,168 @@ func (d *Dialer) Close() error {
...
+func (d *Dialer) pollDomainNames(ctx context.Context) {
+	type cacheEntry struct {
+		cn    instance.ConnName
+		cache monitoredCache
+	}
+
+	// List all the cache entries created with a domain name
+	d.mu.RLock()
+	cacheEntries := make([]cacheEntry, 0, len(d.cache))
+	for cn, cache := range d.cache {
```
enocom, line 603: Looks like we're copying all the keys and values out of this map into a slice. To avoid introducing a new type (cacheEntry), why don't we just allocate a new map of cn -> cache and then iterate over its keys and values? This reduces some of the complexity that adding a new type brings.

dialer.go
enocom: Instead of tracking closed (which requires more locking), could we just always close the connection and ignore double closed errors?

dialer.go
```diff
@@ -538,62 +580,168 @@ func (d *Dialer) Close() error {
...
+			d.mu.Unlock()
+		}
+	}
+
+}
+
+func (d *Dialer) closeDomainNameChanged(ctx context.Context, cn instance.ConnName, cache monitoredCache, err error) {
+	d.removeCached(ctx, cn, cache, err)
+	if atomic.LoadUint64(cache.openConns) > 0 {
+		for _, socket := range cache.openSockets {
```
enocom, line 656: Shouldn't we also reset openSockets here too?

dialer.go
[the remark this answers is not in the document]
enocom: Looking at this code some more, I think monitoredCache is pretty much what we're looking for. The challenge is establishing a clean interface between the dialer (which tracks all the caches) and the monitored cache itself.

I see a few places below where you could move code into methods on monitoredCache.

enocom (commented):
This looks better, but I think there are still some interface problems. I've left detailed comments below.

hessjcg (commented):
I responded to comments. I also significantly restructured the code. Please review #865 first, and then return to this one.

hessjcg:
@enocom Please review #865 before you take another look at this one.

enocom:
To summarize our conversation:

1. We'll move the ticker onto the monitoredCache
2. The monitoredCache will need to have some state when the DNS changes
3. New calls to get a connectionInfoCache will check for that state to see if the dialer needs to throw out the existing cache

hessjcg:
Ok, I've updated the code to reflect our discussion.


I this as discussed: I removed the "lock the world" operations. I moved the domain name ticker to monitoredCache. I made the cache get operation check if an entry is closed.

I still think is valuable to separate the concerns: dialing connections, maintaining a consistent and thread-safe cache state into two classes. `Dialer` should do the dialing, `dialerCache` should maintain cache thread safety and implement the logic around keeping the cache consistent when there are entries with the same domain name but different instance names, or when there are entries that were closed and need to be replaced.

hessjcg:
This is superseded by by #868

PR #850, opened 2024-08-02
feat: support Cloud SQL CAS instances.

feng-zhe:
The CAS instances will have a different way to verify the server identity. This CL supports that.


feng-zhe:
FYI, manually tested the auth proxy client with this PR against the CAS and non-CAS instances. It works well.

jackwotherspoon:
Thanks @feng-zhe I'll take a deep look at this early next week 😄

internal/cloudsql/refresh_test.go
```diff
@@ -79,24 +81,65 @@ func TestRefresh(t *testing.T) {
 	if wantPrivateIP != gotIP {
 		t.Fatalf("metadata IP mismatch, want = %v, got = %v", wantPrivateIP, gotIP)
 	}
-	gotPSC, ok := rr.addrs[PSC]
+	gotDNS, ok := rr.addrs[PSC]
```
jackwotherspoon, line 84: this looks like a typo...
```suggestion
	gotPSC, ok := rr.addrs[PSC]
```

feng-zhe: Thanks, Jack. I just renamed it to `gotPSCDNS` to make it clear that the DNS is from PSC addresses instead of the direct DNS name of the connection info `rr`.

Fixing the failure introduced by the new test now. Thanks.

internal/cloudsql/instance.go
```diff
@@ -171,7 +171,10 @@ type ConnectionInfo struct {
 	ConnectionName    instance.ConnName
 	ClientCertificate tls.Certificate
 	ServerCaCert      *x509.Certificate
+	ServerCaCertPem   []byte // Only set for CAS instances.
```
enocom, line 174: Let's make this ServerCACertPEM (and the above should be ServerCACert). Ditto below wherever we use CA (i.e., "CA" not "Ca", "PEM", not "Pem").

internal/cloudsql/instance.go
```diff
@@ -225,6 +234,16 @@ func (c ConnectionInfo) Addr(ipType string) (string, error) {
 // TLSConfig constructs a TLS configuration for the given connection info.
 func (c ConnectionInfo) TLSConfig() *tls.Config {
 	pool := x509.NewCertPool()
+	if c.ServerCaMode == "GOOGLE_MANAGED_CAS_CA" {
+		pool.AppendCertsFromPEM(c.ServerCaCertPem)
+		return &tls.Config{
+			ServerName:         c.DNSName,
+			Certificates:       []tls.Certificate{c.ClientCertificate},
+			RootCAs:            pool,
+			InsecureSkipVerify: false,
```
enocom, line 243: This line can be omitted and we can just used the zero value.

enocom (changes requested):
Could we add an e2e test for this feature as well? I assume we'd need a new Cloud SQL Instance with this CA mode configured correctly. @jackwotherspoon can you help with that?

internal/cloudsql/instance.go
hessjcg: We probably want have a single field here `ServerCaCert []*x509.Certificate`, instead of two fields populated differently depending on ServerCaMode

internal/cloudsql/instance.go
```diff
@@ -171,7 +171,10 @@ type ConnectionInfo struct {
 	ConnectionName    instance.ConnName
 	ClientCertificate tls.Certificate
 	ServerCaCert      *x509.Certificate
+	ServerCaCertPem   []byte // Only set for CAS instances.
+	ServerCaMode      string // Only set for CAS instances.
```
hessjcg, line 175: `ServerCAMode` (Server Certificate Authority Mode)

internal/cloudsql/instance.go
```diff
@@ -171,7 +171,10 @@ type ConnectionInfo struct {
...
 	DBVersion         string
+	DNSName           string // Only set for CAS instances.
```
hessjcg, line 177: `DomainName` (not DNSName)

internal/cloudsql/instance.go
```diff
@@ -180,18 +183,24 @@ type ConnectionInfo struct {
 // NewConnectionInfo initializes a ConnectionInfo struct.
 func NewConnectionInfo(
 	cn instance.ConnName,
+	dnsName string,
```
hessjcg, line 186: Incorporate names and structure comments from above.

internal/cloudsql/refresh.go
```diff
@@ -47,9 +47,12 @@ const (
 // metadata contains information about a Cloud SQL instance needed to create
 // connections.
 type metadata struct {
-	ipAddrs      map[string]string
-	serverCaCert *x509.Certificate
-	version      string
+	ipAddrs         map[string]string
+	serverCaCert    *x509.Certificate
```
hessjcg, line 51: Same as above, use a single field here instead of two fields: `serverCaCert` and `serverCaCertPem`

internal/cloudsql/instance.go
```diff
@@ -225,6 +234,16 @@ func (c ConnectionInfo) Addr(ipType string) (string, error) {
...
+		return &tls.Config{
+			ServerName:         c.DNSName,
+			Certificates:       []tls.Certificate{c.ClientCertificate},
+			RootCAs:            pool,
+			InsecureSkipVerify: false,
+			MinVersion:         tls.VersionTLS13,
+		}
+	}
 	pool.AddCert(c.ServerCaCert)
 	return &tls.Config{
```
hessjcg, line 248: You will need to modify the `verifyPeerCertificateFunc()`. On like 287, it checks the certificate CommonName field against the expected instance name. This check should not occur when a custom SAN is in use.

hessjcg (changes requested):
Nice start. This needs some restructuring to avoid redundant fields.

hessjcg:
You may also want to incorporate the changes from #843. This will allow customers to configure the connector using a DNS name.

internal/cloudsql/instance.go
enocom: Agreed -- if we can do this with one field, that's best.

feng-zhe: Thanks. Used the `ServerCACert []*x509.Certificate` only now.

internal/cloudsql/instance.go
feng-zhe: Thanks. Done. Also updated other references.

internal/cloudsql/instance.go
feng-zhe: I am not sure about this. Because this `DNSName` is directly from the `dnsName` of the `ConnectSettings` API [here](https://cloud.google.com/sql/docs/mysql/admin-api/rest/v1beta4/connect/get#response-body).

internal/cloudsql/instance.go
feng-zhe: Thanks. Done, except the `dnsName` which I am not sure.

internal/cloudsql/instance.go
feng-zhe: SG. Just removed it. Thanks.

internal/cloudsql/instance.go
feng-zhe: I believe for CAS instances we can rely on the default verification logic since the `InsecureSkipVerify` is false and we set the `ServerName` to the DNS name. We don't have to call `verifyPeerCertificateFunc()`.

internal/cloudsql/refresh.go
feng-zhe: SG. I used the `serverCACert` now.

feng-zhe:
> You may also want to incorporate the changes from #843. This will allow customers to configure the connector using a DNS name.

Unfortunately, for CAS instances, there will be no DNS records for it.  The only change would be the server CA certificates , the DNS name, and how we verify it. I feel #843 is a separate feature where we do have SRV records set up. WDYT? Thanks.

internal/cloudsql/instance.go
enocom: Probably better to stick with the wording from the ConnectSettings API. We could add a godoc comment explaining what this is used for though.

feng-zhe:
> Could we add an e2e test for this feature as well? I assume we'd need a new Cloud SQL Instance with this CA mode configured correctly. @jackwotherspoon can you help with that?

Thanks. Jack helped me and I have added some e2e tests for it. However, the testing PR #854 failed due to some password issue. Debugging it now. Thanks.

feng-zhe:
@enocom Hi Eno, FYI, with @jackwotherspoon's help, the added e2e test has passed in #854. Thanks.

enocom (approved):
A big milestone. Thanks Henry! :clap:

jackwotherspoon (approved):
LGTM ✅ Thanks a ton Henry!

hessjcg (approved):
LGTM. Nice work.

PR #856, opened 2024-08-08
feat: add `bytes_sent` and `bytes_received` as metrics

jackwotherspoon:
Adding additional metrics to be reported by Go Connector.

The Connector will now report the number of `bytes_sent` to Cloud SQL
and `bytes_received` from Cloud SQL.

### DIFF
--- a/internal/trace/metrics.go
+++ b/internal/trace/metrics.go
@@ -104,6 +104,20 @@ var (
 		Aggregation: view.Count(),
 		TagKeys:     []tag.Key{keyInstance, keyDialerID, keyErrorCode},
 	}
+	bytesSentView = &view.View{
+		Name:        "cloudsqlconn//bytes_sent_count",
+		Measure:     mBytesSent,
+		Description: "The number of bytes sent to Cloud SQL",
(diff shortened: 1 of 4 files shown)


internal/trace/metrics.go
```diff
@@ -94,6 +104,20 @@ var (
 		Aggregation: view.Count(),
 		TagKeys:     []tag.Key{keyInstance, keyDialerID, keyErrorCode},
 	}
+	bytesSentView = &view.View{
+		Name:        "cloudsqlconn//bytes_sent",
```
enocom, line 108: Double slash here.

enocom: Ditto below.

enocom (commented):
Shall we add a test in https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/blob/main/metrics_test.go for this?

jackwotherspoon:
Screenshot of metrics appearing in Cloud Monitoring:

`bytes_sent`:
image

`bytes_received`:
image

enocom (approved):

PR #860, opened 2024-08-13
feat: add support for Go 1.23 and drop Go 1.20

jackwotherspoon:
Go 1.20 is EOL: https://endoflife.date/go

Go 1.23 was just released


enocom (approved):
We'll need to update to Go 1.23 in the Proxy as well.

Also, this needs a go mod tidy, and you'll need to remove the stale checks for Go 1.20 and Go 1.22.

PR #865, opened 2024-09-10
chore: Refactor dialer cache concurrency logic. Part of #842.

hessjcg:
This moves concurrency code for the map of instance names to connectionInfoCache out of `Dialer` 
and into a new class `dialerCache`. 

This also updates all uses of `monitoredCache` to use a reference rather than a value. The code needs 
to pass instances of `monitoredCache`  by reference before `monitoredCache` can hold thread-safe
logic managing an individual cache entry.

### DIFF
--- a/dialer.go
+++ b/dialer.go
@@ -113,17 +113,40 @@ type connectionInfoCache interface {
 // monitoredCache is a wrapper around a connectionInfoCache that tracks the
 // number of connections to the associated instance.
 type monitoredCache struct {
-	openConns *uint64
+	openConnsCount *uint64
 
 	connectionInfoCache
+}
+
+func (c *monitoredCache) Close() error {
+	return c.connectionInfoCache.Close()
+}
+
+func (c *monitoredCache) ForceRefresh() {
+	if c == nil || c.connectionInfoCache == nil {
+		return
+	}
+	c.connectionInfoCache.ForceRefresh()
+}
+
+func (c *monitoredCache) UpdateRefresh(b *bool) {
+	if c == nil || c.connectionInfoCache == nil {
+		return
+	}
+	c.connectionInfoCache.UpdateRefresh(b)
+}
+func (c *monitoredCache) ConnectionInfo(ctx context.Context) (cloudsql.ConnectionInfo, error) {
+	if c == nil || c.connectionInfoCache == nil {
+		return cloudsql.ConnectionInfo{}, nil
+	}
+	return c.connectionInfoCache.ConnectionInfo(ctx)
 }
 
 // A Dialer is used to create connections to Cloud SQL instances.
 //
 // Use NewDialer to initialize a Dialer.
 type Dialer struct {
-	lock           sync.RWMutex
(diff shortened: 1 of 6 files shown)


dialer.go
```diff
@@ -538,11 +564,11 @@ func (d *Dialer) Close() error {
 	default:
 	}
 	close(d.closed)
-	d.lock.Lock()
-	defer d.lock.Unlock()
-	for _, i := range d.cache {
-		i.Close()
-	}
+
+	d.cache.replaceAll(func(_ instance.ConnName, c *monitoredCache) (instance.ConnName, *monitoredCache) {
```
enocom, line 568: What will this replaceAll function do? Right now I see it's not using any closure state.

dialer.go
```diff
@@ -551,47 +577,52 @@ func (d *Dialer) Close() error {
...
-				)
-			}
-			var count uint64
-			c = monitoredCache{openConns: &count, connectionInfoCache: cache}
-			d.cache[cn] = c
-		}
+) (*monitoredCache, error) {
+
+	c, oldC, err := d.cache.getOrAdd(cn, func() (*monitoredCache, error) {
+		return d.createConnectionInfoCache(ctx, cn, useIAMAuthN)
```
enocom, line 583: This function calls UpdateRefresh and then we have a second call below. Can we avoid that? Is it intentional?

dialer_cache.go
```diff
@@ -0,0 +1,170 @@
...
+			return cn, c, true
+		}
+	}
+	return instance.ConnName{}, nil, false
+}
+
+// get returns the instance matching the cn.
+//
+// This method is not re-entrant.
+func (d *dialerCache) get(cn instance.ConnName) (*monitoredCache, bool) {
```
enocom, line 91: https://go.dev/doc/effective_go#Getters

enocom (commented):
Some quick comments.

dialer_cache.go
hessjcg: This is an indexed accessor like `cache[key]`, not a getter. Other well-known go cache libraries have chosen to name their "get from cache" method `Get(key)`, so we have done the same here.

- [istio Cache](https://pkg.go.dev/istio.io/pkg/cache#Cache)
- [bigcache](https://github.com/allegro/bigcache)
- [go-cache](https://github.com/patrickmn/go-cache)

hessjcg (commented):
Updated based on review from @enocom.

dialer.go
```diff
@@ -551,47 +577,52 @@ func (d *Dialer) Close() error {
...
-					d.dialerID, useIAMAuthNDial,
-				)
-			}
-			var count uint64
-			c = monitoredCache{openConns: &count, connectionInfoCache: cache}
-			d.cache[cn] = c
-		}
+) (*monitoredCache, error) {
+
+	c, oldC, err := d.cache.getOrAdd(cn, func() (*monitoredCache, error) {
```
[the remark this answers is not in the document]

hessjcg:
This is superseded by by #868

PR #868, opened 2024-09-18
feat: Automatically reset connection when the DNS record changes.

hessjcg:
When a connection is configured using a DNS domain name, the connector will poll the DNS TXT record every 30 seconds. If the value of the DNS record changes, the connector will close all connections to the old instance, and direct new connections to the updated instance. 

Part of https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/issues/842

### DIFF
--- a/README.md
+++ b/README.md
@@ -307,13 +307,12 @@ zone has a TXT record with the value `my-project:region:my-instance`. The
 application establishes connections to the `my-project:region:my-instance` 
 Cloud SQL instance. 
 
-Then, to reconfigure the application using a different database
-instance: `my-project:other-region:my-instance-2`. You update the DNS record
-for `prod-db.mycompany.example.com` with the target 
-`my-project:other-region:my-instance-2`
+Then, to reconfigure the application to use a different database
+instance, change the value of the `prod-db.mycompany.example.com` DNS record
+from `my-project:region:my-instance` to `my-project:other-region:my-instance-2`
 
 The connector inside the application detects the change to this
-DNS entry. Now, when the application connects to its database using the 
+DNS record. Now, when the application connects to its database using the 
 domain name `prod-db.mycompany.example.com`, it will connect to the
 `my-project:other-region:my-instance-2` Cloud SQL instance. 
 
--- a/README.md
+++ b/README.md
@@ -234,7 +234,8 @@ func connect() {
     // ... etc
 }
 ```
-### Using DNS to identify an instance
+
+### Using DNS domain names to identify instances
 
 The connector can be configured to use DNS to look up an instance. This would
 allow you to configure your application to connect to a database instance, and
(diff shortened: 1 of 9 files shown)


hessjcg:
This supersedes #844 and #865.

hessjcg:
Based on our discussion, I have updated `Dialer.cache` to use a cache key that identifies an entry either by "project:region:instance" or by "domainName", not the combination "project:region:instance" + "domainName". I also simplified the logic in `Dialer.connectionInfoCache()` so that it no longer checks if a domain name value is stale before returning existing cache entries.

monitored_cache.go
```diff
@@ -0,0 +1,151 @@
...
+	closedCh         chan struct{}
+
+	mu        sync.Mutex
+	openConns []*instrumentedConn
+	closed    bool
+
+	connectionInfoCache
+}
+
+func newMonitoredCache(ctx context.Context, cache connectionInfoCache, cn instance.ConnName, failoverPeriod time.Duration, resolver instance.ConnectionNameResolver, logger debug.ContextLogger) *monitoredCache {
```
enocom, line 48: nit: wrapping this by stacking the arguments someway would make this nice for people on smaller screens (or splitting code).

monitored_cache.go
```diff
@@ -0,0 +1,151 @@
...
+				_ = socket.Close() // force socket closed, ok to ignore error.
+			}
+		}
+		atomic.StoreUint64(c.openConnsCount, 0)
+	}
+
+	return c.connectionInfoCache.Close()
+}
+
+func (c *monitoredCache) ForceRefresh() {
```
enocom, line 107: By embedding the interface above, you get these delegators automatically, so as long as you're just calling through, you don't need these methods.

README.md
```diff
@@ -292,6 +293,41 @@ func connect() {
...
+will be directed to the new instance. 
+
+For example: suppose application is configured to connect using the
+domain name `prod-db.mycompany.example.com`. Initially the corporate DNS 
+zone has a TXT record with the value `my-project:region:my-instance`. The
+application establishes connections to the `my-project:region:my-instance` 
+Cloud SQL instance. 
+
+Then, to reconfigure the application using a different database
+instance: `my-project:other-region:my-instance-2`. You update the DNS record
```
enocom, line 311: I think there should be a comma in this line, as it's not a complete sentence otherwise.

Then, to reconfigure ..., you update the ...

enocom (approved):
LGTM

jackwotherspoon (approved):
LGTM 👏

dialer.go
```diff
@@ -546,55 +575,86 @@ func (d *Dialer) Close() error {
...
+
+	// c exists and is not closed
+	if ok && !c.isClosed() {
+		c.UpdateRefresh(useIAMAuthN)
+		return c, nil
+	}
+
+	// c exists and is closed, remove it from the cache
+	if ok {
+		delete(d.cache, k)
```
enocom, line 625: I think you can delete this since we re-assign over the key below. cc @jackwotherspoon

monitored_cache.go
hessjcg: Fixed.

PR #871, opened 2024-09-19
test: Add e2e test to access instance using DNS TXT record. part of #842

hessjcg:
Add e2e test to verify that connecting with a DNS name works correctly.

### DIFF
--- a/e2e_mysql_test.go
+++ b/e2e_mysql_test.go
@@ -15,12 +15,15 @@
 package cloudsqlconn_test
 
 import (
+	"context"
 	"database/sql"
+	"fmt"
 	"os"
 	"testing"
 	"time"
 
 	"cloud.google.com/go/cloudsqlconn"
+	"cloud.google.com/go/cloudsqlconn/instance"
 	"cloud.google.com/go/cloudsqlconn/mysql/mysql"
 	gomysql "github.com/go-sql-driver/mysql"
 )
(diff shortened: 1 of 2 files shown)


e2e_mysql_test.go
```diff
@@ -54,25 +57,52 @@ func requireMySQLVars(t *testing.T) {
 	}
 }
 
+type mockResolver struct {
+}
+
+func (r *mockResolver) Resolve(_ context.Context, name string) (instanceName instance.ConnName, err error) {
+	if name == "mysql.example.com" {
+		return instance.ParseConnNameWithDomainName(mysqlConnName, "mysql.example.com")
```
jackwotherspoon, line 65: is this what we would expect a global DNS format to look like? Or would it be in the form with the trailing dot similar to PSC?

hessjcg: The domain name is customer-defined and customer-entered, so I'd expect it to omit the trailing dot.

e2e_mysql_test.go
```diff
@@ -85,18 +115,18 @@ func TestMySQLDriver(t *testing.T) {
...
 			if err != nil {
 				t.Fatalf("failed to register driver: %v", err)
 			}
 			defer cleanup()
 			cfg := gomysql.NewConfig()
 			cfg.CheckConnLiveness = true
-			cfg.User = mysqlUser
-			cfg.Passwd = mysqlPass
+			cfg.User = tc.user
+			cfg.Passwd = tc.password
```
jackwotherspoon, line 126: why do we need to be changing the user and pass per test case?

hessjcg: The test should use the user/password mysqlUser and mysqlPassword for normal connections. The test should use mysqlIAMUser and "password" when the test includes the option WithIAMAuthN().

jackwotherspoon: It looks like the MySQL IAM authN test wasn't even doing and IAM AuthN login previously then... as `mysqlIAMUser` was unused from what I can see?

jackwotherspoon (approved):
LGTM 👍

PR #902, opened 2024-12-19
fix: DNS Resolver should include the domain name in the ConnName.

hessjcg:
The DNS Resolver should return a ConnName with the resolved instance name and the DNS name.


jackwotherspoon (approved):
LGTM, nice catch

PR #910, opened 2025-01-09
chore: Simplify logic choosing legacy or standard server cert validation 

hessjcg:
Going forward, both GOOGLE_MANAGED_CAS_CA, CUSTOMER_MANAGED_CAS_CA, and future new kinds 
of CA will use standard TLS domain name validation using the server certificate SAN records. The certificate
validation logic for the original GOOGLE_MANAGED_INTERNAL_CA is now the exception. 

See implementation in other connectors:
- https://github.com/GoogleCloudPlatform/cloud-sql-nodejs-connector/pull/408
- https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/pull/2095


jackwotherspoon (approved):
LGTM ✅

PR #911, opened 2025-01-09
feat: Validate the server certificate with the domain name from config 

hessjcg:
When the connector is configured using a DNS name (instead of an instance connection name), validate the
instance server certificate using that DNS name. The server certificate should include a Subject Alternative Name
record matching the DNS name used by the connector.


jackwotherspoon (changes requested):
Need a test for this new behavior 🧪

PR #920, opened 2025-01-22
fix: correct default usage for `WithQuotaProject` and `WithUserAgent`

### LINKED ISSUE #918: `WithQuotaProject` conflicts with `WithHTTPClient` in default code path
As part of https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/pull/909 the default usage of the Go Connector leverages the new `auth` library to generate the client.

https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/blob/3f3aebd9df5f9c8dc0830579806045af34a92367/dialer.go#L257-L261

It seems this is causing conflicts as `WithHTTPClient` and `WithQuotaProject` do not work together and results in the following error:

```
... failed to create sqladmin client: WithHTTPClient is incompatible with QuotaProject
```

Should set quota on client if `cloudsqlconn.WithQuotaProject` option is used.

jackwotherspoon:
In #909 we started leveraging [`httptransport.NewClient`](https://pkg.go.dev/cloud.google.com/go/auth/httptransport#NewClient) with [`Option.WithHTTPClient`](https://pkg.go.dev/google.golang.org/api/option#WithHTTPClient).

https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/blob/42d001979ffbe9e5f1323b54eace20a4eb23dd7a/dialer.go#L250-L260

However `WithHTTPClient` can not be used with `WithQuotaProject` or with `WithUserAgent` Options.

If auth client is being used, we must set the headers manually.

Fixes #918

### DIFF
--- a/dialer.go
+++ b/dialer.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"net/http"
 	"os"
 	"strings"
 	"sync"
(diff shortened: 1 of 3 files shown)


jackwotherspoon:
Will add a test to verify and stop future regression

PR #934, opened 2025-01-31
fix: Refresh client cert when CA rotation

### LINKED ISSUE #932: Refresh proxy certificates if the client certificate becomes invalid
### Bug Description

When the server CA certificates are rotated using the [Rotate CA Certificate](https://cloud.google.com/sql/docs/mysql/manage-ssl-instance#rotate-server-certificate-cas) operation, the proxy should recognize that the server certificates are no longer valid, and do a refresh operation. 

### Example code (or command)

```Go

```

### Stacktrace

```bash

```

### Steps to reproduce?

1. ?
2. ?
3. ?
...

### Environment

1. OS type and version:
2. Go version:
3. Cloud SQL Go Connector version:

### Additional Details

_No response_

hessjcg:
When client CA is rotated, this can cause TLS read errors after the `Dialer.Dial()` has returned. The certificate should be refreshed. 

This adds logic to the dialer to refresh the connection cache if a TLS error occurs after `net.Conn` is returned to the database driver. 

Fixes #932

### DIFF
--- a/dialer.go
+++ b/dialer.go
@@ -419,9 +419,11 @@ func (d *Dialer) Dial(ctx context.Context, icn string, opts ...DialOption) (conn
 	tlsConn := tls.Client(conn, ci.TLSConfig())
 	err = tlsConn.HandshakeContext(ctx)
 	if err != nil {
+		// TLS handshake errors are fatal and require a refresh. Remove the instance
+		// from the cache so that future calls to Dial() will block until the
+		// certificate is refreshed successfully.
 		d.logger.Debugf(ctx, "[%v] TLS handshake failed: %v", cn.String(), err)
-		// refresh the instance info in case it caused the handshake failure
-		c.ForceRefresh()
+		d.removeCached(ctx, cn, c, err)
 		_ = tlsConn.Close() // best effort close attempt
 		return nil, errtype.NewDialError("handshake failed", cn.String(), err)
 	}
@@ -433,10 +435,26 @@ func (d *Dialer) Dial(ctx context.Context, icn string, opts ...DialOption) (conn
 		trace.RecordDialLatency(ctx, icn, d.dialerID, latency)
 	}()
 
-	iConn := newInstrumentedConn(tlsConn, func() {
+	closeFunc := func() {
 		n := atomic.AddUint64(c.openConnsCount, ^uint64(0)) // c.openConnsCount = c.openConnsCount - 1
 		trace.RecordOpenConnections(context.Background(), int64(n), d.dialerID, cn.String())
-	}, d.dialerID, cn.String())
+	}
+	errFunc := func(err error) {
+		// io.EOF occurs when the server closes the connection. This is safe to
+		// ignore.
+		if err == io.EOF {
+			return
+		}
+		d.logger.Debugf(ctx, "[%v] IO Error on Read or Write: %v", cn.String(), err)
+		if d.isTLSError(err) {
+			// TLS handshake errors are fatal. Remove the instance from the cache
+			// so that future calls to Dial() will block until the certificate
+			// is refreshed successfully.
+			d.removeCached(ctx, cn, c, err)
+			_ = tlsConn.Close() // best effort close attempt
+		}
+	}
+	iConn := newInstrumentedConn(tlsConn, closeFunc, errFunc, d.dialerID, cn.String())
 
 	// If this connection was opened using a Domain Name, then store it for later
 	// in case it needs to be forcibly closed.
@@ -447,12 +465,19 @@ func (d *Dialer) Dial(ctx context.Context, icn string, opts ...DialOption) (conn
 	}
 	return iConn, nil
 }
+func (d *Dialer) isTLSError(err error) bool {
+	if nErr, ok := err.(net.Error); ok {
+		return !nErr.Timeout() && // it's a permanent net error
+			strings.Contains(nErr.Error(), "tls") // it's a TLS-related error
+	}
+	return false
+}
 
-// removeCached stops all background refreshes and deletes the connection
-// info cache from the map of caches.
+// removeCached stops all background refreshes, closes open sockets, and deletes
+// the cache entry.
 func (d *Dialer) removeCached(
 	ctx context.Context,
-	i instance.ConnName, c connectionInfoCache, err error,
+	i instance.ConnName, c *monitoredCache, err error,
 ) {
 	d.logger.Debugf(
 		ctx,
@@ -460,10 +485,19 @@ func (d *Dialer) removeCached(
 		i.String(),
 		err,
 	)
+
+	// If this instance of monitoredCache is still in the cache, remove it.
+	// If this instance was already removed from the cache or
+	// if *a separate goroutine* replaced it with a new instance, do nothing.
+	key := createKey(i)
 	d.lock.Lock()
-	defer d.lock.Unlock()
+	if cachedC, ok := d.cache[key]; ok && cachedC == c {
+		delete(d.cache, key)
+	}
+	d.lock.Unlock()
+
+	// Close the monitoredCache, this call is idempotent.
 	c.Close()
-	delete(d.cache, createKey(i))
 }
 
 // validClientCert checks that the ephemeral client certificate retrieved from
@@ -505,7 +539,7 @@ func (d *Dialer) EngineVersion(ctx context.Context, icn string) (string, error) 
 	}
 	ci, err := c.ConnectionInfo(ctx)
 	if err != nil {
-		d.removeCached(ctx, cn, c.connectionInfoCache, err)
+		d.removeCached(ctx, cn, c, err)
 		return "", err
 	}
 	return ci.DBVersion, nil
@@ -529,17 +563,18 @@ func (d *Dialer) Warmup(ctx context.Context, icn string, opts ...DialOption) err
 	}
 	_, err = c.ConnectionInfo(ctx)
 	if err != nil {
-		d.removeCached(ctx, cn, c.connectionInfoCache, err)
+		d.removeCached(ctx, cn, c, err)
 	}
 	return err
 }
 
 // newInstrumentedConn initializes an instrumentedConn that on closing will
 // decrement the number of open connects and record the result.
-func newInstrumentedConn(conn net.Conn, closeFunc func(), dialerID, connName string) *instrumentedConn {
+func newInstrumentedConn(conn net.Conn, closeFunc func(), errFunc func(error), dialerID, connName string) *instrumentedConn {
 	return &instrumentedConn{
 		Conn:      conn,
 		closeFunc: closeFunc,
+		errFunc:   errFunc,
 		dialerID:  dialerID,
 		connName:  connName,
 	}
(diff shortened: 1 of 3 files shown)


dialer.go
```diff
@@ -461,8 +485,11 @@ func (d *Dialer) removeCached(
 	)
 	d.lock.Lock()
 	defer d.lock.Unlock()
-	c.Close()
-	delete(d.cache, createKey(i))
+	key := createKey(i)
+	if cachedC, ok := d.cache[key]; ok && cachedC == c {
+		delete(d.cache, key)
+	}
```
jackwotherspoon, line 491: nit: can we add a comment around this logic?

What case is this for?

dialer_test.go
```diff
@@ -1174,3 +1174,134 @@ func TestDialerChecksSubjectAlternativeNameAndFails(t *testing.T) {
...
+			mock.InstanceGetSuccess(inst, 2),
+			mock.CreateEphemeralSuccess(inst, 2),
+		},
+		dialerOptions: []Option{
+			WithTokenSource(mock.EmptyTokenSource{}),
+			WithDebugLogger(&dialerTestLogger{t: t}),
+			//WithLazyRefresh(),
+			// Note: this succeeds with lazy refresh, but fails with lazy.
+			// because dialer.ForceRefresh does not block connections while the
+			// refresh is in progress.
```
jackwotherspoon, line 1198: I think there is a typo in this comment...?

jackwotherspoon (changes requested):

dialer.go
```diff
@@ -435,7 +437,22 @@ func (d *Dialer) Dial(ctx context.Context, icn string, opts ...DialOption) (conn
 	iConn := newInstrumentedConn(tlsConn, func() {
 		n := atomic.AddUint64(c.openConnsCount, ^uint64(0)) // c.openConnsCount = c.openConnsCount - 1
 		trace.RecordOpenConnections(context.Background(), int64(n), d.dialerID, cn.String())
-	}, d.dialerID, cn.String())
+	},
+		func(err error) {
+			// ignore EOF
```
enocom, line 442: Could we expand this comment on why we're ignoring EOF? I think EOFs are the server sending an RST packet, but that's not obvious knowledge for future maintainers.

dialer.go
```diff
@@ -435,7 +437,22 @@ func (d *Dialer) Dial(ctx context.Context, icn string, opts ...DialOption) (conn
 	iConn := newInstrumentedConn(tlsConn, func() {
 		n := atomic.AddUint64(c.openConnsCount, ^uint64(0)) // c.openConnsCount = c.openConnsCount - 1
 		trace.RecordOpenConnections(context.Background(), int64(n), d.dialerID, cn.String())
-	}, d.dialerID, cn.String())
+	},
+		func(err error) {
```
enocom, line 441: whitespace nit: putting this up a line would make it more obvious it's an argument at a glance.

Alternatively, you could pull this anonymous function out into a variable:

``` golang
errFunc := func(err error) {
  // ...
}
iConn := newInstrumentedConn(/*...*/, errFunc)
```

dialer.go
```diff
@@ -446,12 +463,19 @@ func (d *Dialer) Dial(ctx context.Context, icn string, opts ...DialOption) (conn
 	}
 	return iConn, nil
 }
+func (d *Dialer) isTLSError(err error) bool {
+	if nErr, ok := err.(net.Error); ok {
+		return !nErr.Timeout() && // it's a permanent net error
```
enocom, line 468: Have you manually tested this logic as part of doing a cert rotation to confirm it's working as intended?

dialer_test.go
enocom: +1 This looks like a dev-artifact that isn't true anymore?

dialer_test.go
```diff
@@ -1174,3 +1174,134 @@ func TestDialerChecksSubjectAlternativeNameAndFails(t *testing.T) {
...
+	t.Log("Second attempt should fail...")
+	conn, err := d.Dial(context.Background(), "my-project:my-region:my-instance")
+	if err != nil {
+		t.Fatal("Should be no certificate error after, got ", err)
+	}
+
+	// Expect an error on read. This should trigger the dialer to refresh.
+	_, err = io.ReadAll(conn)
+	if err != nil {
+		t.Log("Got error on read as expected.", err)
```
enocom, line 1228: t.Logf("Got error on read as expected: %v")

dialer_test.go
```diff
@@ -1174,3 +1174,134 @@ func TestDialerChecksSubjectAlternativeNameAndFails(t *testing.T) {
...
+
+	// Start the server with new certificates
+	cancel2 := mock.StartServerProxy(t, inst)
+	defer cancel2()
+
+	// Dial a second time. We expect no error on dial, but TLS error on read.
+	t.Log("Second attempt should fail...")
+	_, err := d.Dial(context.Background(), "my-project:my-region:my-instance")
+	if err != nil {
+		t.Log("Got error on dial as expected.", err)
```
enocom, line 1287: s/Got/got/

dialer_test.go
```diff
@@ -1174,3 +1174,134 @@ func TestDialerChecksSubjectAlternativeNameAndFails(t *testing.T) {
...
+	} else {
+		t.Fatal("Want dial error, got no error")
```
enocom, line 1289: s/Want/want/

dialer_test.go
```diff
@@ -1174,3 +1174,134 @@ func TestDialerChecksSubjectAlternativeNameAndFails(t *testing.T) {
...
+	// Dial again. This should complete after the refresh.
+	t.Log("Third attempt...")
+	testSuccessfulDial(
+		context.Background(), t, d,
+		"my-project:my-region:my-instance",
+	)
+	t.Log("Third attempt OK.")
+}
+
+func TestDialerRefreshesAfterRotateServerCA(t *testing.T) {
```
enocom, line 1243: Could this test and the one above be combined into a table test by parameterizing the only differences between the two tests?

internal/mock/cloudsql.go
```diff
@@ -75,11 +75,13 @@ func (f FakeCSQLInstance) serverCACert() ([]byte, error) {
 	if f.signer != nil {
 		return f.signer(f.Cert, f.Key)
 	}
+
 	if f.serverCAMode == "" || f.serverCAMode == "GOOGLE_MANAGED_INTERNAL_CA" {
 		// legacy server mode, return only the server cert
 		return toPEMFormat(f.certs.serverCert)
 	}
 	return toPEMFormat(f.certs.casServerCertificate, f.certs.serverIntermediateCaCert, f.certs.serverCaCert)
+
```
enocom, line 84: nit: let's omit these changes from the PR since they're unrelated

internal/mock/sqladmin.go
```diff
@@ -102,38 +102,38 @@ func (r *Request) matches(hR *http.Request) bool {
...
-		PscEnabled:      i.pscEnabled,
-		ServerCaMode:    i.serverCAMode,
-	}
-
 	r := &Request{
 		reqMethod: http.MethodGet,
 		reqPath:   fmt.Sprintf("/sql/v1beta4/projects/%s/instances/%s/connectSettings", i.project, i.name),
 		reqCt:     ct,
 		handle: func(resp http.ResponseWriter, _ *http.Request) {
+			var ips []*sqladmin.IpMapping
```
enocom, line 110: Is this a necessary change? Looks like formatting to me?

dialer_test.go
hessjcg: Fixed.

dialer_test.go
hessjcg: Test parameterized.

dialer_test.go
hessjcg: Done.

dialer_test.go
hessjcg: Done.

internal/mock/cloudsql.go
hessjcg: Fixed.

internal/mock/sqladmin.go
hessjcg: Yes, this is necessary. The Server CA changes during the test. The ServerCaCert response field needs to hold the up-to-date value. Thus, this method needs to create the response at the time that the request is made.

dialer.go
enocom: Might be worth adding an e2e test for this, but otherwise this seems fine.

dialer.go
```diff
@@ -447,23 +465,39 @@ func (d *Dialer) Dial(ctx context.Context, icn string, opts ...DialOption) (conn
...
+	i instance.ConnName, c *monitoredCache, err error,
 ) {
 	d.logger.Debugf(
 		ctx,
 		"[%v] Removing connection info from cache: %v",
 		i.String(),
 		err,
 	)
+
+	// If this instance of monitoredCache is still in the cache, remove it.
```
enocom, line 489: Would we just always call delete, regardless of whether the monitoredCache is present or not in cache?

dialer_test.go
```diff
@@ -1174,3 +1174,126 @@ func TestDialerChecksSubjectAlternativeNameAndFails(t *testing.T) {
...
 }
+
+func TestDialerRefreshesAfterRotateCACerts(t *testing.T) {
+	tcs := []struct {
+		desc            string
+		rotateClientCA  bool
+		wantErrorOnDial bool
+		wantErrorOnRead bool
+		useLazyRefresh  bool
+	}{{
```
enocom, line 1185: nit: this isn't consistently formatted as below

dialer_test.go
```diff
@@ -1174,3 +1174,126 @@ func TestDialerChecksSubjectAlternativeNameAndFails(t *testing.T) {
...
+		t.Run(tc.desc, func(t *testing.T) {
+			inst := mock.NewFakeCSQLInstanceWithSan(
+				"my-project", "my-region", "my-instance", []string{"db.example.com"},
+				mock.WithDNS("db.example.com"),
+				mock.WithServerCAMode("GOOGLE_MANAGED_CAS_CA"),
+			)
+
+			opts := []Option{
+				WithTokenSource(mock.EmptyTokenSource{}),
+				//WithDebugLogger(&dialerTestLogger{t: t}),
```
enocom, line 1217: Can this be deleted?

internal/mock/sqladmin.go
enocom: Got it -- a comment to that effect above `i.serverCACert()` would help future maintainers.

dialer_test.go
hessjcg: I uncommented it. It helps to understand the test.

dialer.go
enocom: Ah got it. I missed that detail. I agree then -- no e2e is fine.

dialer.go
enocom: I see. That's probably an edge case, but also worth solving for. Maybe add a note like, "in the case of multiple goroutines creating cache entries, only purge the cache if it's for a known failed attempt and don't overwrite separate goroutine entries." This is a subtle little detail otherwise.

enocom: Or maybe just:

``` golang
// If this instance was already removed from the cache or *a separate goroutine* replaced it with
// a new instance, do nothing.
```

enocom (approved):

jackwotherspoon (approved):
LGTM ✅ 

Thanks for the extra assist here @enocom 👏

PR #947, opened 2025-02-14
chore: throw unused errors

jackwotherspoon:
The `err` for `certDerBytes, err := x509.CreateCertificate(rand.Reader, cert, cert, &k.PublicKey, k)` 
was being unused.

Adding check for error and throwing proper error in tests.


jackwotherspoon:
better option would be to pass the `*testing.T` parameter through and fail the test instead of panic'ing

kgala2 (approved):

PR #948, opened 2025-02-19
chore: audit csql test instances

kgala2:
Consolidate use of Cloud SQL instances for end to end tests in the following cases:
1. MySQL IAM 
2. Postgres IAM

### DIFF
--- a/e2e_mysql_test.go
+++ b/e2e_mysql_test.go
@@ -32,7 +32,7 @@ var (
 	// "Cloud SQL MySQL instance connection name, in the form of 'project:region:instance'.
 	mysqlConnName = os.Getenv("MYSQL_CONNECTION_NAME")
 	// "Cloud SQL MySQL instance connection name, in the form of 'project:region:instance'.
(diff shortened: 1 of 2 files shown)


e2e_mysql_test.go
```diff
@@ -32,7 +32,7 @@ var (
 	// "Cloud SQL MySQL instance connection name, in the form of 'project:region:instance'.
 	mysqlConnName = os.Getenv("MYSQL_CONNECTION_NAME")
 	// "Cloud SQL MySQL instance connection name, in the form of 'project:region:instance'.
-	mysqlIAMConnName = os.Getenv("MYSQL_IAM_CONNECTION_NAME")
+	mysqlIAMConnName = os.Getenv("MYSQL_CONNECTION_NAME")
```
jackwotherspoon, line 35: nit: should we not just get rid of `mysqlIAMConnName` as it will be identical to `mysqlConnName`?

jackwotherspoon (changes requested):

jackwotherspoon (approved):

PR #957, opened 2025-03-18
chore: remove toolchain from go.mod

jackwotherspoon:
Using the Go toolchain is really only needed when you support old versions of Go.

We only support active versions, same as the Go language does. 

Toolchain is thus not needed.


jackwotherspoon:
Closing, as #954 removes toolchain

PR #965, opened 2025-04-14
chore: update IP Type options in Test PG Auth

kgala2:
This PR updates call to AddIPTypeOptions function to correctly add PrivateIP option based on ipType

### DIFF
--- a/e2e_postgres_test.go
+++ b/e2e_postgres_test.go
@@ -738,8 +738,8 @@ func TestPostgresAuthentication(t *testing.T) {
 	var opts []cloudsqlconn.Option
 	if ipType != "private" {
 		creds = keyfile(t)
-		opts = AddIPTypeOptions(opts)
 	}
+	opts = AddIPTypeOptions(opts)
 	tok, path, cleanup := removeAuthEnvVar(t)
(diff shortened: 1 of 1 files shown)


e2e_postgres_test.go
```diff
@@ -738,8 +738,8 @@ func TestPostgresAuthentication(t *testing.T) {
 	var opts []cloudsqlconn.Option
 	if ipType != "private" {
 		creds = keyfile(t)
-		opts = AddIPTypeOptions(opts)
 	}
+	opts = AddIPTypeOptions(opts)
```
jackwotherspoon, line 742: nit: this should be called `addIpTypeOptions`, in Go you only capitalize the first letter of a func if you want it to be exported, I am pretty sure this func is only used within this file (i.e. is internal)?

```suggestion
	opts = addIpTypeOptions(opts)
```

jackwotherspoon (approved):
See my nit: can be a follow-up refactor PR

e2e_postgres_test.go
kgala2: Noted, thanks!

PR #983, opened 2025-05-17
fix: reduce the number of memory allocations and the latency overhead.

dominiquelefevre:
Struct instrumentedConn would trace every call to Read() and Write(). This incurs a significant overhead because the variadic arguments of trace.RecordBytesReceived() escape and need to be heap-allocated.

Also, every trace.RecordBytesReceived() would be called in a new goroutine. I think this was done to decrease the latency overhead. I am not certain it achieves that goal, but it certainly introduces a race. Different calls to trace.RecordBytesReceived() can race with each other and decrease the counter of received bytes.

It makes no sense to update the performance counter this often. The default scraping interval in Prometheus is 1 minute. Google Cloud Monitoring had the same interval before they added high-resolution counters that scrape services every 10 seconds.

Only update integer counters in the hot path, and update OpenCensus' counters once in 5 seconds.

I have a test that just loops through `SELECT * FROM t WHERE id = $1` and has response sizes that range from several dozen bytes to several kilobytes. At 64 connections to Postgres and 10k requests per connection, the test makes approx. 25.5M allocations before this patch, and 7.2M allocations after the patch. The CPU time goes down accordingly because OpenCensus and the garbage collector have less work to do.


dialer.go
```diff
@@ -571,13 +571,21 @@ func (d *Dialer) Warmup(ctx context.Context, icn string, opts ...DialOption) err
 // newInstrumentedConn initializes an instrumentedConn that on closing will
 // decrement the number of open connects and record the result.
 func newInstrumentedConn(conn net.Conn, closeFunc func(), errFunc func(error), dialerID, connName string) *instrumentedConn {
-	return &instrumentedConn{
+	reporterCtx, stopReporter := context.WithCancel(context.Background())
```
enocom, line 574: ctx, cancel would be fine here.

https://google.github.io/styleguide/go/decisions#variable-names

dialer.go
```diff
@@ -571,13 +571,21 @@ func (d *Dialer) Warmup(ctx context.Context, icn string, opts ...DialOption) err
...
 		closeFunc: closeFunc,
 		errFunc:   errFunc,
 		dialerID:  dialerID,
 		connName:  connName,
+
+		reportTicker: time.NewTicker(5 * time.Second),
+		stopReporter: stopReporter,
 	}
+
+	go c.reportCountersGor(reporterCtx)
```
enocom, line 586: Shorter is OK here too:

```
go c.report(ctx)
```

dialer.go
```diff
@@ -590,14 +598,19 @@ type instrumentedConn struct {
...
+	stopReporter func()
 }
 
 // Read delegates to the underlying net.Conn interface and records number of
 // bytes read
 func (i *instrumentedConn) Read(b []byte) (int, error) {
 	bytesRead, err := i.Conn.Read(b)
 	if err == nil {
-		go trace.RecordBytesReceived(context.Background(), int64(bytesRead), i.connName, i.dialerID)
+		atomic.AddInt64(&i.bytesRead, int64(bytesRead))
```
enocom, line 613: Shall we use https://pkg.go.dev/sync/atomic#Int64.Add instead here?

enocom: Ditto below.

dialer.go
```diff
@@ -571,13 +571,21 @@ func (d *Dialer) Warmup(ctx context.Context, icn string, opts ...DialOption) err
...
 func newInstrumentedConn(conn net.Conn, closeFunc func(), errFunc func(error), dialerID, connName string) *instrumentedConn {
-	return &instrumentedConn{
+	reporterCtx, stopReporter := context.WithCancel(context.Background())
+	c := &instrumentedConn{
 		Conn:      conn,
 		closeFunc: closeFunc,
 		errFunc:   errFunc,
 		dialerID:  dialerID,
 		connName:  connName,
+
```
enocom, line 581: unnecessary new line

internal/trace/metrics.go
```diff
@@ -108,14 +108,14 @@ var (
 		Name:        "cloudsqlconn/bytes_sent",
 		Measure:     mBytesSent,
 		Description: "The number of bytes sent to Cloud SQL",
-		Aggregation: view.Sum(),
+		Aggregation: view.LastValue(),
```
enocom, line 111: This is a breaking change. Perhaps we could keep `Sum` and instead use https://pkg.go.dev/sync/atomic#Int64.Swap in the reporting goroutine to reset the byte counter to 0? What do you think?

enocom (commented):
Thanks for this PR.

I'm surprised by the arguments to RecordBytesReceived escaping the stack. Do you understand why this is happening?

In any case, you're right that we don't need to schedule goroutines with every read and write. So this is a big improvement. To avoid a breaking change in the metrics, I think we need to make a small adjustment to the code here. See below for my suggestion.

Thanks again for sending this.

internal/trace/metrics.go
dominiquelefevre: AFAIUI, `Sum()` and `LastValue()` decide how OpenCensus handles values provided to `stats.Record()`. They do not change the output that Prometheus receives. Prometheus still gets one number that is the total number of bytes sent or received over a connections since it was established.

dominiquelefevre:
> I'm surprised by the arguments to RecordBytesReceived escaping the stack. Do you understand why this is happening?

The lifetime of a goroutine is not limited to the lifetime of a stack frame where it was created, so all arguments of a goroutine have to be heap-allocated. The compiler allocates one object to hold all parameters, but it still has to allocate.

Another problem is the use `context.WithValue()` and, pointers-to-functions and interface methods.
```
func RecordBytesReceived(ctx context.Context, num int64, instance, dialerID string) {
    // tag.New() assumes that output parameters escape so ctx is always heap-allocated.
    // context.WithValue() puts a reference to tags into ctx, so tags escape as well.
    ctx, _ = tag.New(ctx, tag.Upsert(keyInstance, instance), tag.Upsert(keyDialerID, dialerID))

    // stats.Record() forwards its arguments to internal.MeasurementRecorder.(measurementRecorder).
    // It is not a function, but a pointer to a function. The compiler assumes that all arguments escape,
    // in particular, mBytesReceived.M(num). The same is true for all interface methods.
    stats.Record(ctx, mBytesReceived.M(num))
}
```

internal/trace/metrics.go
enocom: For people who export to Cloud Monitoring, this setting controls how the initial metric descriptor is created. Sum maps to a cumulative metric, and last value maps to a gauge metric. Changing the aggregation will cause any additional writes to the metric to fail.

I think we can use use atomic.SwapInt64 to capture the existing count and reset the count to zero. And then we avoid the breaking change here.

``` golang
func (i *instrumentedConn) reportCounters() {
  bytesRead := atomic.SwapInt64(&i.bytesRead, 0)
  bytesWritten := atomic.SwapInt64(&i.bytesWritten, 0)
  trace.RecordBytesReceived(context.Background(), bytesRead, i.connName, i.dialerID)
  trace.RecordBytesSent(context.Background(), bytesWritten, i.connName, i.dialerID)
}
```

dialer.go
enocom: There's apparently a bug around AddInt64 for 32-bit platforms, but we can address this in a separate PR.

internal/trace/metrics.go
dominiquelefevre: Oh wow :/

Ok, then `Sum() + SwapInt64()` it is.

enocom (approved):
Really appreciate this PR. This is a big win for anyone using the Go Connector.

enocom:
@hessjcg Would you like to take a pass on this as well now that I'm technically not a code owner anymore?

dominiquelefevre:
I will have time to fix the tests like `TestMonitoredCache_Close` next week. If you help me with those and merge the PR earlier, I will be happy.

enocom:
@hessjcg looks like the test just needs to properly initialize an instrumentedConn based on the changes here.

dominiquelefevre:
Fixed `TestDialerWithMetrics` and `TestMonitoredCache_Close`.

enocom:
See https://github.com/GoogleCloudPlatform/alloydb-go-connector/pull/686 where we ported this change and used `(*atomic.Int64) Add` to avoid alignment problems in 32 bit architectures.

dominiquelefevre:
Hey @hessjcg! Is anyone out here to read PRs?

hessjcg (approved):
Thank you for this improvement. This looks great!

dominiquelefevre:
@hessjcg please make a release so that I can pull in this fix without referencing some random git commit.

chuan-z:
hi @dominiquelefevre , thanks for this great contribution! We have monthly release scheduled. And we will release latest with this PR included in next 1 ~ 2 weeks.

PR #999, opened 2025-08-11
chore: Update to golangci-lint v2.3.1

hessjcg:
Updates the golangci lint configuration and build files to use v2.3.1

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -50,24 +50,47 @@ function e2e() {
   e2e_ci
 }
 
+# e2e_ci - Run end-to-end integration tests in the CI system.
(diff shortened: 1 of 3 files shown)


build.sh
```diff
@@ -63,11 +65,14 @@ function fix() {
 ## lint - runs the linters
 function lint() {
   # run golangci-lint
+  golangci_lint_version=v2.3.1
```
kgala2, line 68: can we fetch the latest version of golangci runtime? or do we want it to be 2.3.1 for now?

kgala2 (approved):

Issue #936, opened 2025-02-04
Incorrect log message for usage with domain names


jackwotherspoon:
There is buggy logic around a `Dial` log message:
https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/blob/fef405bd5c4893b8c6b2edd5fe7a4e24142b839f/dialer.go#L348-L350

When `Dial` is called with a DNS domain name as the `icn` this logic message code path will always be executed (even if the domain name still points at the same instance) even when it shouldn't be.

The reason being that `cn.String()` has additional formatting done to it:
https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/blob/fef405bd5c4893b8c6b2edd5fe7a4e24142b839f/instance/conn_name.go#L41-L44

Example: If a user calls `Dial` with the domain name `example.com` as the `icn`. 
Then `cn.String()` would always return `example.com -> my-project:region:instance`
and continuously log a debug message.

hessjcg:
This is a good starter issue.

panavenue:
Issued fixed.

PR #1006, opened 2025-08-20
fix: fix logmessage for usage with domain names

panavenue:
issue: #936 

Instead of using the original string() method which contains the dns name. Update it to follow the standard icn name standard.

### DIFF
--- a/dialer.go
+++ b/dialer.go
@@ -350,7 +350,7 @@ func (d *Dialer) Dial(ctx context.Context, icn string, opts ...DialOption) (conn
 	// Log if resolver changed the instance name input string.
 	if cn.DomainName() != "" {
 		// icn is a domain name, which resolves to a actual icn
-		d.logger.Debugf(ctx, "resolved dns: %s", icn, cn.String())
+		d.logger.Debugf(ctx, "resolved domain name %s to %s", icn, cn.String())
 	} else if cn.String() != icn {
 		// icn was not a domain name, but the resolver changed it and cn != icn
 		d.logger.Debugf(ctx, "resolved instance connection string %s to %s", icn, cn.String())


hessjcg:
Maybe we could do something like this:
```go
	// Log if resolver changed the instance name input string.
	// icn = "pro:reg:inst" -> cn = "pro:reg:inst" cn.String() = "pro:reg:inst"
	// icn = "example.com" -> cn = "example.com", "pro:reg:inst" cn.String() = "example.com -> pro:reg:inst"
	// icn = "my-custom-name" -> cn = "pro:reg:inst" cn.String() = "pro:reg:inst"
	if cn.DomainName != "" {
	    // icn is a domain name
		d.logger.Debugf(ctx, "resolved instance domain name %s to %s", icn, ... pro:reg:inst)
	}
	else if cn.String() != icn {
	    // icn was not a domain name, but the resolver changed it and cn != icn
		d.logger.Debugf(ctx, "resolved instance connection string %s to %s", icn, cn)
	}
```

panavenue:
> Maybe we could do something like this:
> 
> ```go
> 	// Log if resolver changed the instance name input string.
> 	// icn = "pro:reg:inst" -> cn = "pro:reg:inst" cn.String() = "pro:reg:inst"
> 	// icn = "example.com" -> cn = "example.com", "pro:reg:inst" cn.String() = "example.com -> pro:reg:inst"
> 	// icn = "my-custom-name" -> cn = "pro:reg:inst" cn.String() = "pro:reg:inst"
> 	if cn.DomainName != "" {
> 	    // icn is a domain name
> 		d.logger.Debugf(ctx, "resolved instance domain name %s to %s", icn, ... pro:reg:inst)
> 	}
> 	else if cn.String() != icn {
> 	    // icn was not a domain name, but the resolver changed it and cn != icn
> 		d.logger.Debugf(ctx, "resolved instance connection string %s to %s", icn, cn)
> 	}
> ```

Good call! I think I get the logic now. 

I modified a bit for the `cn.DomainName != ""` part, because `cn.String()` already in a correct format of "DomainName -> ICN"

PR #1011, opened 2025-09-08
feat: Add support for the Metadata Exchange protocol

hessjcg:
This is the client-side implementation of the metadata exchange protocol for Cloud SQL instances. This will
allow connectors to pass metadata about the connection to the Cloud SQL instance while the connection
is being established. 

In particular, this is a part of resolving the issues with the Auth Proxy connecting to MySQL instances that
use the caching_sha2_password extension. (See https://github.com/GoogleCloudPlatform/cloud-sql-proxy/issues/2317)


internal/cloudsql/metadataexchange.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+
+	// Get the MDX message length from the MDX header
+	msgLen := binary.BigEndian.Uint32(mdxHeaderBytes[mdxSignatureLength:mdxHeaderLength])
+	if msgLen > maxMessageSize {
+		return nil, errtype.NewDialError(fmt.Sprintf("MDX Read Format: Response too long, %d bytes", msgLen), c.cn, err)
+	}
+
+	// Read the MDX message
+	mdxBytes := make([]byte, msgLen)
```
enocom, line 148: To avoid the performance tax of heap allocation, we use a sync.Pool in the AlloyDB Go Connector.

See https://github.com/GoogleCloudPlatform/alloydb-go-connector/blob/227c2d92167aa9c62ed904588ce202f71f37fc7c/dialer.go#L501-L502 and https://github.com/GoogleCloudPlatform/alloydb-go-connector/blob/227c2d92167aa9c62ed904588ce202f71f37fc7c/dialer.go#L554-L577

panavenue (approved):

PR #1026, opened 2025-10-09
fix: downgrade go version back to 1.24

panavenue:


panavenue:
Closing this now. No need to downgrade because Go patched a good version.

Issue #1032, opened 2025-10-18
Panic in internal/cloudsql/refresh.go


synaptor:
### Bug Description

When I recreated my CloudSQL Postgres database, previously running cloud-sql-proxy crashed with a stack trace pointing to this library.

### Example code (or command)

```Go
cloud-sql-proxy --port 6000 --auto-iam-authn ...databasename...
```

### Stacktrace

```bash
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xac82c2]
 
goroutine 866 [running]:
cloud.google.com/go/cloudsqlconn/internal/cloudsql.fetchMetadata({0x13c4ea8?, 0xc000b90000?}, 0xc0007e07e0, {{0xc000416000, 0x14}, {0xc0000100b0, 0xd}, {0xc000010140, 0x4}, {0x0, ...}})
        /root/go/pkg/mod/cloud.google.com/go/cloudsqlconn@v1.18.1/internal/cloudsql/refresh.go:134 +0x982
cloud.google.com/go/cloudsqlconn/internal/cloudsql.adminAPIClient.ConnectionInfo.func2()
        /root/go/pkg/mod/cloud.google.com/go/cloudsqlconn@v1.18.1/internal/cloudsql/refresh.go:304 +0xc5
created by cloud.google.com/go/cloudsqlconn/internal/cloudsql.adminAPIClient.ConnectionInfo in goroutine 865
        /root/go/pkg/mod/cloud.google.com/go/cloudsqlconn@v1.18.1/internal/cloudsql/refresh.go:302 +0x345
```

### Steps to reproduce?

1. Run cloud-sql-proxy
2. Open a connection to the database
3. Delete the database
4. Create the database
5. Observe crash

### Environment

1. OS type and version: Linux Debian
2. Go version: 1.23.4
3. Cloud SQL Go Connector version: 1.18.1

kgala2:
Hi @synaptor,

Looks like the instance's CA certificates are not fully propagated when the proxy tries to connect to it. To verify this can you please check if the instance is fully created and in a runnable state in the Google Cloud Console before the proxy attempts to connect?

Could you also check whether using the cloud-sql-proxy now (when the instance is in runnable state) facing any panics?

synaptor:
[...]

PR #1037, opened 2025-10-23
chore: add checks for server ca certificates in refresh.go

kgala2:
This change addresses a panic that can occur during the refresh operation if the server CA certificate is not available in the instance metadata.

A nil check is added to validate the server CA certificate before it is used. If the certificate is nil, a refresh error is now returned instead of causing a panic from a nil pointer dereference.

Fixes: #1032


enocom (approved):
LGTM.

From an API perspective, is this an acceptable state?

kgala2:
> LGTM.
> 
> From an API perspective, is this an acceptable state?

The API does not return server CA cert while creation (the first minute of it)

panavenue (approved):

PR #1054, opened 2025-12-16
feat: Use configured DNS name to lookup instance IP address

### LINKED ISSUE #1053: Use configured DNS name to lookup instance IP address
## Feature Description

When a connector is configured using a DNS name, it uses the DNS name to look up the instance connection name in a TXT record. The connector should use the DNS name to look up the instance IP address when connecting. 

This is particularly useful for PSC instances with custom DNS names. By adding TXT and A records for the instance, the connector will only use the custom DNS name for the instance. It will not use the Cloud SQL-assigned PSC DNS name. 

Users must create and update the TXT and A DNS records themselves. These records are not managed by Cloud SQL. 

If the connector cannot resolve the IP address for the configured DNS name, it will fall back to existing behavior: use the IP address or cloud-sql assigned PSC DNS name from the instance metadata.

## Updated Connection Algorithm

This is how a connector can connect to an instance with a custom DNS name AND an A record with the instance's IP address: 

GIVEN:

- The custom DNS name for an instance: "db.example.com"
- An instance "my-project:us-central1:my-db" 
- The instance is configured with the Custom SAN "db.example.com"
- DNS TXT record "db.example.com" that resolves to "my-project:us-central1:my-db" 
- DNS A record "db.example.com" that resolves to the IP address of the instance

WHEN: the connector dials the instance

THEN:
- Connector performs DNS lookup of the  TXT record for the instance to get the instance connection name "my-project:us-central1:my-db" 
- Connector calls the SqlAdmin API to get the instance metadata
- Connector performs DNS lookup of the A record for the instance to get the instance IP address
- Connector opens TCP connection to {instance-ip}, port 3307
- Connector performs mTLS handshake with hostname validation for "db.example.com" 
- Connector dial operation is complete.

If there is no A or AAAA record for the DNS name address, it should fall back to its current behavior: using the IP address or the Cloud SQL-assigned PSC DNS name from the instance metadata. 

GIVEN:

- The custom DNS name for an instance: "db.example.com"
- An instance "my-project:us-central1:my-db" 
- The instance is configured with the Custom SAN "db.example.com"
- DNS TXT record "db.example.com" that resolves to "my-project:us-central1:my-db" 
- There is no A record defined for "db.example.com"

WHEN: the connector dials the instance

THEN:
- Connector performs DNS lookup of the  TXT record for the instance to get the instance connection name "my-project:us-central1:my-db" 
- Connector calls the SqlAdmin API to get the instance metadata
- Connector performs DNS lookup of the A record for the instance to get the instance IP address, no a record is found
- Connector uses the instance metadata to determine the instance IP address
- Connector opens TCP connection to {instance-ip}, port 3307
- Connector performs mTLS handshake with hostname validation for "db.example.com" 
- Connector dial operation is complete.

hessjcg:
When a custom DNS name is used to connect to a Cloud SQL instance, the dialer should first attempt to resolve the custom DNS name to an IP address and use that for the connection. If the lookup fails, the dialer should fall back to using the IP address from the instance metadata.

This change modifies the dialer to:
- Use the configured resolver to look up the host's IP address.
- Use the IP address from the A record of the custom DNS name if available.
- Fall back to the IP address from the instance metadata if the A record is not available.

Fixes #1053

### DIFF
--- a/dialer_test.go
+++ b/dialer_test.go
@@ -1514,8 +1514,8 @@ func TestDialerChecksSubjectAlternativeNameAndFallsBackToCN(t *testing.T) {
 	d := setupDialer(t, setupConfig{
 		testInstance: inst,
 		reqs: []*mock.Request{
-			mock.InstanceGetSuccess(inst, 1),
-			mock.CreateEphemeralSuccess(inst, 1),
+			mock.InstanceGetSuccess(inst, 2),
+			mock.CreateEphemeralSuccess(inst, 2),
 		},
 
 		dialerOptions: []Option{
@@ -1526,8 +1526,7 @@ func TestDialerChecksSubjectAlternativeNameAndFallsBackToCN(t *testing.T) {
 			}),
 			WithTokenSource(mock.EmptyTokenSource{}),
 			WithDNSResolver(),
-		},
-	})
+		}})
 
 	tcs := []struct {
 		desc string


hessjcg:
I have fixed the test failures in `TestDialerChecksSubjectAlternativeNameAndFallsBackToCN`. The test was failing because it performed two dial operations (one using the instance connection name and another using the domain name), which triggered two separate metadata requests due to different cache keys. The mock, however, was configured to expect only one request. I updated the test to expect two requests, and verified that the tests now pass locally. The changes have been pushed to the branch.

hessjcg:
I have resolved the linting issues in `dialer_test.go` by fixing the indentation. I verified that the tests still pass and pushed the changes. The PR should be ready for review now.

hessjcg:
I have verified that the linting checks pass and the tests are green. The PR is now fully updated with the fixes.

dialer.go
```diff
@@ -407,6 +410,28 @@ func (d *Dialer) Dial(ctx context.Context, icn string, opts ...DialOption) (conn
...
+				"[%v] custom DNS name %q did not resolve to an IP address: %v, using %s from instance metadata",
+				cn.String(), cn.DomainName(), err, addr)
+		} else if len(addrs) == 0 {
+			d.logger.Debugf(ctx,
+				"[%v] custom DNS name %q resolved but returned no entries, using %s from instance metadata",
+				cn.String(), cn.DomainName(), addr)
+		} else {
+			d.logger.Debugf(ctx,
+				"[%v] custom DNS name %q resolved to %q, using it to connect",
+				cn.String(), cn.DomainName(), addrs[0])
```
panavenue: I am guessing it is always in the first item of the array?

hessjcg: Usually it only returns 1 result. If it returns more than 1 result, that is legal, but probably a mistake.

PR #1085, opened 2026-04-10
chore: deprecate pgv4

panavenue:
Deprecate pgv4, due to the fact that https://github.com/jackc/pgproto3 is End of Life and no longer supported.


examples/cloudrun/mysql/go.mod
```diff
@@ -1,42 +1,41 @@
 module cloud.google.com/go-cloudrun/mysql
 
-go 1.24.0
-
-toolchain go1.24.3
+go 1.25.7
```
kgala2: do we want to update this to 1.25.7?

panavenue: Good question. 

I think it was automatically updated when I ran `go get cloud.google.com/go/cloudsqlconn@latest`

since our go connector requires go version 1.25.7 now - https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/blob/main/go.mod#L3

enocom (commented):
You can solve this problem without a breaking change like this: https://github.com/GoogleCloudPlatform/alloydb-go-connector/pull/767

examples/cloudrun/mysql/go.mod
hessjcg: Yes. This should match the go version in the connector's go.mod

hessjcg:
> You can solve this problem without a breaking change like this: [GoogleCloudPlatform/alloydb-go-connector#767](https://github.com/GoogleCloudPlatform/alloydb-go-connector/pull/767)

+1 to this approach. We can remove the dependency on pgxv4 without breaking our public API.

panavenue:
> You can solve this problem without a breaking change like this: [GoogleCloudPlatform/alloydb-go-connector#767](https://github.com/GoogleCloudPlatform/alloydb-go-connector/pull/767)

Good call!

panavenue:
> > You can solve this problem without a breaking change like this: [GoogleCloudPlatform/alloydb-go-connector#767](https://github.com/GoogleCloudPlatform/alloydb-go-connector/pull/767)
> 
> +1 to this approach. We can remove the dependency on pgxv4 without breaking our public API.

Yeah this is good suggestion. Thank you @hessjcg . 

Btw, since we are not breaking our API with this approach. We might have to reconsider Major vs Minor upgrade for our release.

examples/cloudrun/mysql/go.mod
enocom: The go version should match the lowest version the Connector supports which is 1.25.0.

The toolchain meanwhile can be whatever you're using to develop on.

enocom: Also this should be a separate PR (traditionally we do feat: make go xxx the minimum version)

enocom (commented):
This should be a fix commit in my opinion so you can alert users to the change.

examples/cloudrun/mysql/go.mod
hessjcg: +1. Please split this into a separate PR.

hessjcg (approved):
With the caveat that the go version change goes in a separate PR.

examples/cloudrun/mysql/go.mod
panavenue: Reverted back the version, and will create a new PR for it. 

But I am not sure if this will succee. I believe `go 1.25.7` is needed for the example, since it need to match the version we have in mod

panavenue: Nvm, looks like that was fine, I will have a separate PR for this.

PR #1106, opened 2026-05-29
feat: Add PSC DNS and Global Write Endpoint support to Go Connector

hessjcg:
The Private Service Connect (PSC) DNS Name Resolution & Fallback feature enables the Go connector to securely connect to Cloud SQL database instances using Private Service Connect (PSC) DNS, custom domain names (CNAMEs), and Global Write Endpoints.

The connector can now dial PSC instances using:
- Its regional DNS hostname (e.g., `0123456789ab.fedcba9876543.us-central1.sql-psc.goog`)
- A custom domain CNAME pointing to the instance PSC DNS hostname
- A Global Write Endpoint DNS hostname (e.g., `0123456789ab.fedcba9876543.global.sql-psc.goog`)

The connectors will dynamically resolve this hostname to its real Cloud SQL instance connection name via the SQL Admin API.

For Global Write Endpoints (which use the "global" region in their DNS name), the resolver bypasses direct API resolution (as "global" is not a valid API region) and forces fallback to CNAME resolution. This CNAME should point to the active regional instance DNS hostname, enabling automatic failover.

Furthermore, if local DNS resolution of the hostname fails (e.g., on local developer networks lacking a private DNS route), the connectors gracefully fall back to dialing the instance's Private or Public IP address dynamically retrieved from metadata.


hessjcg:
This is waiting on the next release of the Cloud SQL API libraries which will include the new ResolveConnectSettings method.

internal/cloudsql/resolver.go
```diff
@@ -133,3 +179,47 @@ func (r *DNSInstanceConnectionNameResolver) queryDNS(ctx context.Context, domain
 	// No records were found, return an error.
 	return instance.ConnName{}, fmt.Errorf("no valid TXT records found for %q", domainName)
 }
+
+// parseDNSName parses a DNS name into its constituent parts.
+// Instance DNS names follow this template:
+// {instance-dns-label}.{project-dns-label}.{cloud-region}.{dns-suffix}
+// Suffix is one of: sql.goog, sql-psa.goog, sql-psc.goog (with optional trailing dot).
+func parseDNSName(dnsName string) (instanceLabel, projectLabel, region, suffix string, ok bool) {
```
hessjcg: This should distinguish between the global write endpoint and regional instance DNS names.

internal/mock/cloudsql.go
```diff
@@ -300,11 +300,19 @@ func GenerateCertWithCommonName(i FakeCSQLInstance, cn string) []byte {
...
 func StartServerProxy(t *testing.T, i FakeCSQLInstance) func() {
 
-	ln, err := tls.Listen("tcp", ":3307", &tls.Config{
-		Certificates: i.certs.serverChain(i.useStandardTLSValidation),
-		ClientCAs:    i.certs.clientCAPool(),
-		ClientAuth:   tls.RequireAndVerifyClientCert,
-	})
+	var ln net.Listener
+	var err error
+	for attempts := 0; attempts < 50; attempts++ {
```
hessjcg: This is an unrelated change. Allows for errors during mock server startup due to attempting to open a port that is already in use.

internal/mock/cloudsql.go
```diff
@@ -410,11 +418,19 @@ func NewFailoverTestServer(t *testing.T) *FailoverTestServer {
...
+		ln, err = tls.Listen("tcp", ":3307", &tls.Config{
```
hessjcg: Same. Unrelated change.

dialer.go
```diff
@@ -493,45 +495,67 @@ func (d *Dialer) connectInstanceIP(ctx context.Context, cn instance.ConnName, cf
...
-		addrs, err := d.dnsResolver.LookupHost(ctx, cn.DomainName())
-		if err != nil {
-			d.logger.Debugf(ctx,
-				"[%v] custom DNS name %q did not resolve to an IP address: %v, using %s from instance metadata",
-				cn.String(), cn.DomainName(), err, addr)
-		} else if len(addrs) == 0 {
-			d.logger.Debugf(ctx,
-				"[%v] custom DNS name %q resolved but returned no entries, using %s from instance metadata",
-				cn.String(), cn.DomainName(), addr)
+		resolved, err := d.dnsResolver.LookupHost(ctx, cn.DomainName())
```
hessjcg, line 510: Clarify this code. Why is this happening?

options.go
```diff
@@ -65,8 +65,10 @@ type dialerConfig struct {
 	dnsResolver              cloudsql.NetResolver
 	sqlDataDialer            sqldataclient.Dialer
 	sqlDataStreamTimeout     time.Duration
+	useDNSNameResolver       bool
 	// err tracks any dialer options that may have failed.
-	err error
+	err    error
+	client *sqladmin.Service
```
hessjcg, line 71: Does this get used?

internal/cloudsql/resolver.go
```diff
@@ -45,45 +47,89 @@ func (r *ConnNameResolver) Resolve(_ context.Context, icn string) (instanceName
...
 		}
+
+		// Attempt to query a TXT record
+		cn, txtErr = r.queryDNS(ctx, current)
+		if txtErr == nil {
+			return cn, nil
+		}
+
+		// If TXT lookup fails, check CNAME record
+		cname, cnameErr := r.dnsResolver.LookupCNAME(ctx, current)
```
panavenue: From Gemini:


- Root Cause: In Go's standard library ([net.Resolver.LookupCNAME](https://pkg.go.dev/net#Resolver.LookupCNAME)), if a domain name does not have a CNAME record (e.g., a direct hostname or A/AAAA record), LookupCNAME returns host. and err == nil.
- Impact: For any standard domain name without a CNAME alias that fails TXT lookup (e.g. typo in domain name, or host lacking TXT records), cname == current evaluates to true, and the resolver returns "cname loop detected for <domain>".
- Fix: When cname == current, it simply means the CNAME chain has ended (canonical name reached). Return txtErr instead of reporting a loop. To detect true CNAME loops (e.g. A -> B -> A), maintain a visited := make(map[string]bool) set across loop iterations.

internal/cloudsql/resolver.go
```diff
@@ -45,45 +47,89 @@ func (r *ConnNameResolver) Resolve(_ context.Context, icn string) (instanceName
...
+
+		// Check if it matches the well-known DNS pattern directly
+		if _, _, region, _, ok := parseDNSName(current); ok {
+			dnsNameWithDot := current
+			if !strings.HasSuffix(dnsNameWithDot, ".") {
+				dnsNameWithDot += "."
+			}
+			db, err := retry50x(ctx, func(ctx2 context.Context) (*sqladmin.ConnectSettings, error) {
+				return r.client.Connect.Resolve(
+					region, dnsNameWithDot).Context(ctx2).Do()
```
panavenue: - Root Cause: If DNSInstanceConnectionNameResolver is instantiated with a nil client (e.g. constructed directly or in unit tests/custom resolvers), dialing any valid Cloud SQL PSC DNS name (e.g., *.sql-psc.goog) immediately triggers a runtime panic on r.client.Connect.
- Fix: Check if r.client == nil before calling r.client.Connect.Resolve and return a descriptive error.

panavenue (commented):
Added some feedback from Gemini

internal/cloudsql/resolver.go
hessjcg: I updated the resolver.go with better error messages to handle this case correctly.

internal/cloudsql/resolver.go
hessjcg: Done.

internal/cloudsql/resolver.go
hessjcg: Instead of adding this if statement, I updated this so that all usage goes through `NewDNSResolver()`. This ensures that this field is initialized. I also renamed `dnsInstanceConnectionNameResolver` so that it is internal only within the package.

internal/mock/cloudsql.go
hessjcg: Moved this change to https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/pull/1131

dialer.go
hessjcg: Simplified to match the existing behavior.

options.go
hessjcg: Removed.

internal/cloudsql/refresh.go
```diff
@@ -382,3 +383,17 @@ func supportsAutoIAMAuthN(version string) error {
 		return fmt.Errorf("%s does not support Auto IAM DB Authentication", version)
 	}
 }
+
+func sortPSCDNSNames(names []string) {
+	sort.SliceStable(names, func(i, j int) bool {
+		iIsPsc := strings.HasSuffix(names[i], ".sql-psc.goog")
```
kgala2, line 395: could the names[] contain DNS with trailing "." ?

internal/cloudsql/resolver.go
```diff
@@ -45,45 +47,94 @@ func (r *ConnNameResolver) Resolve(_ context.Context, icn string) (instanceName
...
+		return instance.ConnName{}, errtype.NewConfigError(
+			"invalid connection name, expected PROJECT:REGION:INSTANCE "+
+				"format or valid DNS domain name",
+			icn,
+		)
+	}
+
+	current := icn
+	var txtErr error
+	for depth := 0; depth < 10; depth++ {
```
kgala2, line 89: question: is there a possibility of self loop in cname resolution? A->B->C->A ?

hessjcg: Yes. But if that happens the loop will terminate at depth 10.

internal/cloudsql/refresh.go
hessjcg: Fixed.

kgala2 (approved):

PR #1108, opened 2026-06-03
feat: Support Developer Edition connections

hessjcg:
This PR adds support for Developer Edition connections via the SqlDataService. It includes a fallback mechanism to standard IP connections if the SqlDataService is not supported by the instance edition.


internal/cloudsql/lazy_test.go
```diff
@@ -143,7 +143,7 @@ func TestLazyRefreshCacheUpdateRefresh(t *testing.T) {
 	}
 
 	if got := spy.callCount(); got != 0 {
-		t.Fatal("auth.TokenProvider was called, but should not have been")
```
hessjcg, line 146: Rollback. This is not a relevant change.

internal/cloudsql/metadataexchange_test.go
```diff
@@ -82,9 +82,7 @@ func newMDXResponseBytes(data []byte) (*mdx.MetadataExchangeResponse, []byte) {
 }
 
 func newMDXRequestBytes(data []byte) (*mdx.MetadataExchangeRequest, []byte) {
-	req := mdx.MetadataExchangeRequest_builder{
```
hessjcg, line 85: Rollback. This is not a relevant change.

internal/cloudsql/refresh.go
```diff
@@ -286,7 +290,7 @@ type adminAPIClient struct {
 	// key is used to generate the client certificate
 	key    *rsa.PrivateKey
 	client *sqladmin.Service
-	// tp is the TokenProvider used for IAM DB AuthN.
+	// tp is the tokenProvider used for IAM DB AuthN.
```
hessjcg: TokenProvider

fallback_conn.go
```diff
@@ -0,0 +1,152 @@
...
+	if firstReadDone {
+		return conn.Read(b)
+	}
+
+	n, err := conn.Read(b)
+	c.firstReadDone = true
+	if err != nil && c.isFallbackError(err) {
+		return c.reconnectAndRead(b)
+	}
+	c.mu.Lock()
```
hessjcg: This seems like strange concurrency logic. Please review.

internal/sqldata/sql_data_service.proto
```diff
@@ -0,0 +1,283 @@
...
+  // `locations/{location}`, for example: `locations/us-central1`.
+  string location_id = 1;
+
+  // Deprecated: Use the `message` oneof instead. The type of message sent
+  // within the stream.
+  oneof message_type {
+    // Deprecated: Use `start_session` or `continue_session` instead.
+    // Parameters for establishing the connection. MUST be sent as the first
+    // message on the stream.
+    ConnectionSettings connection_settings = 2;
```
DeepeshDamodaran: we don't want to show this deprecated field in the first version of the public api. this may be more work on backend side, but i feel we should do that.

DeepeshDamodaran (commented):
had a comment about the api.

fallback_conn.go
```diff
@@ -0,0 +1,152 @@
...
+	firstReadDone = true
```
panavenue: Hmm, looks like this local var is not being use anymore below, do we still need to set it to true?

panavenue (approved):
There is 1 failing unit test. Probably format too strict?

internal/sqldata/sql_data_service.proto
hessjcg: Removed.

PR #1125, opened 2026-07-09
feat: dynamically detect client telemetry attributes and record dialstatus

panavenue:
Continue the initial setup of the client metric. 

- Dynamically detect ClientRegion, ComputePlatform, and DatabaseEngineType in telemetry Config
- Format ResourceID as [project_name:instance_name]
- Upgrade cached MetricRecorder engine type once connection info DBVersion is retrieved
- Protect detection helper functions with panic recovery safeguards

### DIFF
--- a/internal/tel/tel.go
+++ b/internal/tel/tel.go
@@ -129,27 +129,22 @@ func DetectClientRegion() (region string) {
 			region = "UNKNOWN"
 		}
 	}()
-	if metadata.OnGCE() {
-		zone, err := metadata.Zone()
-		if err == nil && zone != "" {
-			if idx := strings.LastIndex(zone, "-"); idx != -1 {
-				return zone[:idx]
-			}
-			return zone
-		}
+	if !metadata.OnGCE() {
+		return "UNKNOWN"
 	}
-	return "UNKNOWN"
+	zone, err := metadata.Zone()
+	if err != nil || zone == "" {
+		return "UNKNOWN"
+	}
+	if idx := strings.LastIndex(zone, "-"); idx != -1 {
+		return zone[:idx]
+	}
+	return zone
 }
 
 // DetectComputePlatform detects the GCP compute platform environment.
 // Any panic or error fails silently and returns "UNKNOWN".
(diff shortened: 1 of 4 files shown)


kgala2 (approved):

internal/tel/tel.go
```diff
@@ -117,6 +120,73 @@ func AuthTypeValue(iamAuthn bool) string {
 	return "built_in"
 }
 
+// DetectClientRegion returns the client region from GCP metadata if running on GCE/GKE/Cloud Run.
+// Any panic or error fails silently and returns "UNKNOWN".
+func DetectClientRegion() (region string) {
+	// Safeguard against unexpected panics to prevent hard crashes in client applications.
+	defer func() {
+		if r := recover(); r != nil {
```
enocom, line 128: This looks unnecessary to me. Are there any missing checks below that would prevent a panic? Does metadata.Zone panic? I would expect not.

internal/tel/tel.go
```diff
@@ -117,6 +120,73 @@ func AuthTypeValue(iamAuthn bool) string {
...
+			region = "UNKNOWN"
+		}
+	}()
+	if metadata.OnGCE() {
+		zone, err := metadata.Zone()
+		if err == nil && zone != "" {
```
enocom: This is  backwards. https://google.github.io/styleguide/go/decisions.html#indent-error-flow

internal/tel/tel.go
enocom: Ditto below.

enocom (changes requested):

internal/tel/tel.go
panavenue: Yeah I was hesitating when adding this. I was afraid that in future, someone might add additional logic in these methods, which cause a panic. 

Will remove.

panavenue: Removed

internal/tel/tel.go
panavenue: Fixed.

enocom (approved):

PR #1129, opened 2026-07-14
feat: add support for connection_durations and open_connection_count …

panavenue:
## 🎯 Overview
Adds built-in client telemetry support for two new Cloud Monitoring (Stackdriver) metrics:
1. `cloudsql.googleapis.com/client/connector/connection_durations`
2. `cloudsql.googleapis.com/client/connector/open_connection_count`

These metrics provide deeper observability into connector connection lifetimes, total dial attempts, and granular error classifications.

---

### 1. `connection_durations` (`Float64Histogram`)
Records the distribution of established connection durations (in seconds) between connection open and connection close.
* **Unit**: `s`
* **Labels**:
  * `instance_auth_type`: one of `[built_in, iam]`
  * `instance_ip_type`: one of `[public, psa, psc]`

### 2. `open_connection_count` (`Int64Counter`)
Cumulative count of all connection attempts (`+1` on each attempt, regardless of success or failure).
* **Unit**: `1`
* **Labels**:
  * `instance_auth_type`: one of `[built_in, iam]`
  * `instance_ip_type`: one of `[public, psa, psc]`
  * `connection_part`: if status is not success, which part of the connection errored out, one of `[client_to_proxy, proxy_to_server]` (empty string `""` on success)
  * `status`: granular status of the dial attempt, one of:
    * `success`
    * `user_error` (e.g., config errors, malformed instance names)
    * `cache_error`
    * `tcp_error` (e.g., dial TCP, keep-alive failures)
    * `tls_error` (e.g., handshake failures, certificate verification errors)
    * `auth_error` (e.g., IAM authentication, permission denied)
    * `refresh_failed_error` (e.g., API errors during certificate/info refresh)
    * `refresh_expired_error` (e.g., refresh failed because client cert expired)
    * `dns_lookup_error`
    * `unknown_error`

### DIFF
--- a/dialer.go
+++ b/dialer.go
@@ -539,7 +539,9 @@ func (d *Dialer) connectInstanceIP(ctx context.Context, cn instance.ConnName, cf
 	attrs.CacheHit = cacheHit
 	if err != nil {
 		attrs.DialStatus = tel.ConnectError
+		attrs.ConnectionPart, attrs.AttemptStatus = tel.ClassifyDialError(err)
 		mr.RecordConnectLatencies(ctx, attrs, time.Since(startTime).Milliseconds())
+		mr.RecordOpenConnectionCount(ctx, attrs)
 		endInfo(err)
 		return nil, err
 	}
@@ -547,7 +549,9 @@ func (d *Dialer) connectInstanceIP(ctx context.Context, cn instance.ConnName, cf
 	if err != nil {
 		d.removeCached(ctx, cn, c, err)
 		attrs.DialStatus = tel.ConnectError
+		attrs.ConnectionPart, attrs.AttemptStatus = tel.ClassifyDialError(err)
 		mr.RecordConnectLatencies(ctx, attrs, time.Since(startTime).Milliseconds())
+		mr.RecordOpenConnectionCount(ctx, attrs)
 		endInfo(err)
 		return nil, err
 	}
@@ -566,6 +570,10 @@ func (d *Dialer) connectInstanceIP(ctx context.Context, cn instance.ConnName, cf
 		ci, err = c.ConnectionInfo(ctx)
 		if err != nil {
 			d.removeCached(ctx, cn, c, err)
+			attrs.DialStatus = tel.ConnectError
+			attrs.ConnectionPart, attrs.AttemptStatus = tel.ClassifyDialError(err)
+			mr.RecordConnectLatencies(ctx, attrs, time.Since(startTime).Milliseconds())
+			mr.RecordOpenConnectionCount(ctx, attrs)
 			return nil, err
 		}
 	}
@@ -577,6 +585,10 @@ func (d *Dialer) connectInstanceIP(ctx context.Context, cn instance.ConnName, cf
 	addr, err := ci.Addr(cfg.connectionType)
 	if err != nil {
 		d.removeCached(ctx, cn, c, err)
+		attrs.DialStatus = tel.ConnectError
+		attrs.ConnectionPart, attrs.AttemptStatus = tel.ClassifyDialError(err)
+		mr.RecordConnectLatencies(ctx, attrs, time.Since(startTime).Milliseconds())
+		mr.RecordOpenConnectionCount(ctx, attrs)
 		return nil, err
 	}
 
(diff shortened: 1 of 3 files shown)


internal/tel/tel.go
```diff
@@ -191,6 +195,30 @@ type Attributes struct {
...
 	// ipType specifies IP address type of the connection, one of [public, psa, psc].
 	IPAddressType string
+	// ConnectionPart specifies which part of the connection errors out, one of [client_to_proxy, proxy_to_server].
+	ConnectionPart string
+	// AttemptStatus specifies the detailed status of the dial attempt.
+	AttemptStatus string
+}
+
+// ClassifyDialError classifies an error during a connection attempt into connection_part and status labels.
+func ClassifyDialError(err error) (part string, status string) {
```
enocom: nit: named returns are discouraged. Use then only when you need to modify them with a defer statement.

internal/tel/tel.go
```diff
@@ -347,6 +393,28 @@ func (m *metricRecorder) RecordConnectLatencies(ctx context.Context, a Attribute
...
+		metric.WithAttributeSet(attribute.NewSet(
+			attribute.String(authType, AuthTypeValue(a.IAMAuthN)),
+			attribute.String(ipType, a.IPType),
+		)),
+	)
+}
+
+// RecordOpenConnectionCount records connection attempts and error classifications.
+func (m *metricRecorder) RecordOpenConnectionCount(ctx context.Context, a Attributes) {
+	m.mOpenConnectionCount.Add(ctx, 1,
```
enocom, line 408: I've lost track if we've fixed this here or not, but FYI https://github.com/GoogleCloudPlatform/alloydb-go-connector/commit/b444a620f2825dedca13bd5d2ab5ac978c4b9843. There was a bug in AlloyDB's recording of this metric.

dialer.go
```diff
@@ -648,18 +672,24 @@ func (d *Dialer) connectInstanceIP(ctx context.Context, cn instance.ConnName, cf
 	}
 
 	latency := time.Since(startTime).Milliseconds()
+	openTime := time.Now()
 	n := c.openConnsCount.Add(1)
 	trace.RecordOpenConnections(ctx, int64(n), d.dialerID, cn.String())
 	trace.RecordDialLatency(ctx, cn.String(), d.dialerID, latency)
 	attrs.DialStatus = tel.ConnectSuccess
+	attrs.ConnectionPart = ""
```
enocom, line 680: Why are you zeroing out this field here?

internal/tel/tel.go
```diff
@@ -282,14 +312,28 @@ func NewMetricRecorder(ctx context.Context, l debug.ContextLogger, cl *monitorin
 		l.Debugf(ctx, "built-in metrics exporter failed to initialize refresh count metric: %v", err)
 		return NullMetricRecorder{}
 	}
+	mConnectionDuration, err := m.Float64Histogram(connectionDurations, metric.WithUnit("s"))
```
enocom, line 315: Is the unit correct here? I would expect ms.

enocom (commented):
Since we're just merging into a parent branch, the description is less important, but when we get to the final PR, we should make sure we distinguish between "feat" (a user facing feature) and otherwise. For now, there is no user-facing change here and so this isn't a new feature.

dialer.go
panavenue: Yep good question. 

So this connectionPart is only recorded if there is an error. If successful, both [client_to_proxy, proxy_to_server] should be success as well.

internal/tel/tel.go
panavenue: Yeah we uses "s" for the duration. I suspect there might be a connection open up for a long period of time before close.

internal/tel/tel.go
panavenue: Fixed.

internal/tel/tel.go
panavenue: Ah, the open_connection_counts here is actually not the "open_connection" gauge metric, that is a separate one. 

But good call! I can double check this in a separate PR.

hessjcg (approved):

PR #1141, opened 2026-08-06
feat: implement cooldown on ResourceExhausted errors

hessjcg:
When SqlDataService returns a RESOURCE_EXHAUSTED error (indicating the service
is busy), the connector will temporarily enter a cooldown period (default 5s)
for that instance. During this period, subsequent dial attempts will fail
immediately with a ResourceExhaustedError instead of calling the server.

This prevents overloading the busy server.


dialer.go
```diff
@@ -960,3 +1032,49 @@ func isPreconditionFailed(err error) bool {
...
+
+func (c *resourceExhaustedTrackingConn) Write(b []byte) (int, error) {
+	n, err := c.Conn.Write(b)
+	if err != nil && isResourceExhaustedError(err) {
+		c.onResourceExhausted(err)
+	}
+	return n, err
+}
+
+func cooldownBackoff(base time.Duration, attempt int) time.Duration {
```
panavenue, line 1079: Just thinking out loud - Should we set the upper bound or the exp backoff duration?

panavenue (approved):
Left a comment but overall LGTM

dialer.go
hessjcg: Yes. I added this a moment ago. There is a maximum on number of attempts, See dialer.go:480

PR #1146, opened 2026-08-12
chore: Use GITHUB_TOKEN in github API requests to avoid rate limits

hessjcg:
This PR updates the CI build scripts to use the GITHUB_TOKEN when fetching tools from GitHub API, preventing rate-limiting issues on CI runners.


panavenue (approved):
Nice! we need this to avoid CI failure.

